PR #9, opened 2017-09-27
Update README instructions and bazel BUILD file

font:
I started going through the README to build the binaries and container images and ran into some issues fixed here:

1. Update script name
1. Update build and run commands
1. Update bazel BUILD file to allow specifying `project` to upload container image
1. Update generated bazel BUILD file
1. Typos

I also ran into an issue because I was missing the Mercurial `hg` CLI tool due to a [bitbucket dependency for `bitbucket.org/ww/goautoneg`](https://github.com/kubernetes/apiserver/blob/b2a8ad67a002d27c8945573abb80b4be543f2a1f/Godeps/Godeps.json#L10) since we have [k8s.io/apiserver in our WORKSPACE dependency](https://github.com/kubernetes/cluster-registry/blob/94f78f21436ae2a8b24ca0fb8d07001c5b7cd52d/WORKSPACE#L61-L67). Installing Mercurial fixed the problem but I'm thinking that should have been handled for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,19 @@ design](https://docs.google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJ
 
 ## Prerequisites
 
-You must have a recent verzion of [bazel](https://bazel.io) installed. Bazel is
+You must have a recent version of [bazel](https://bazel.io) installed. Bazel is
 the recommended way to build and test the cluster registry. Bazel is designed to
 maintain compatibility with standard Go tooling, but this is not tested on a
 regular basis, and some scripts/tooling in this project are built around Bazel.
 
+You will need the [Mercurial](https://www.mercurial-scm.org/) `hg` CLI command
+installed in order to have Bazel install some of the dependencies.
+
 Before doing any development work, you must (in order, from the repository root
 directory, after cloning):
 
-*   run `update-genfiles.sh`
-*   run `bazel run //:gazelle`
+1.  run `update-codegen.sh`
+1.  run `bazel run //:gazelle`
 
 ## Building crinit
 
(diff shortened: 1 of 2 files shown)


font:
@perotinus @madhusudancs

cmd/clusterregistry/BUILD.bazel
```diff
@@ -11,6 +11,28 @@ docker_push(
     name = "push-clusterregistry-image",
     image = ":clusterregistry-image",
     registry = "gcr.io",
-    repository = "kubetest-157500/clusterregistry",
+    repository = "$(project)/clusterregistry",
     tag = "dev",
 )
+
```
perotinus: This bit should not be checked in, since it's autogenerated: can you revert this?

README.md
```diff
@@ -16,16 +16,16 @@ design](https://docs.google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJ
 
 ## Prerequisites
 
-You must have a recent verzion of [bazel](https://bazel.io) installed. Bazel is
+You must have a recent version of [bazel](https://bazel.io) installed. Bazel is
 the recommended way to build and test the cluster registry. Bazel is designed to
 maintain compatibility with standard Go tooling, but this is not tested on a
 regular basis, and some scripts/tooling in this project are built around Bazel.
```
perotinus: Can you add a note about needing hg installed here? Bazel does not take care of that for you, and I don't think it's expected to take care of this level of tool integration, so it's a prerequisite that must be installed separately.

font: I added a note about `hg`. We can probably assume folks have `git` installed. :)

font:
Updated to address comments.

perotinus:
Thanks, Ivan!

PR #14, opened 2017-10-04
 Remove the need for Mercurial to be installed by depending on an HTTP archive instead.

perotinus:
Tested by removing the hg command from the PATH and running `bazel fetch`.


perotinus:
cc @madhusudancs @font

PR #15, opened 2017-10-04
Uncomment a block that configures the API server with authentication options.

perotinus:
Without this, the API server will not authorize any requests.

cc @font @madhusudancs

### DIFF
--- a/cmd/clusterregistry/app/server.go
+++ b/cmd/clusterregistry/app/server.go
@@ -86,9 +86,9 @@ func NonBlockingRun(s *options.ServerRunOptions, stopCh <-chan struct{}) error {
 	if err := s.SecureServing.ApplyTo(genericConfig); err != nil {
 		return err
 	}
-	//if err := s.Authentication.ApplyTo(genericConfig); err != nil {
-	//	return err
-	//}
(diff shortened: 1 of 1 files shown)


pmorie:
@perotinus This uses the k8s authentication from apimachinery, correct?  In that case, the cluster registry is deployed alongside at least one k8s cluster?

perotinus:
@pmorie I'm not quite sure I follow your question, but I'll try to answer.

The cluster registry uses the k8s authentication approach and libraries. It currently is set up to enable a large set of the authentication methods that a standard k8s cluster API server will accept, though I'm not sure it's enabled exactly the same ones. It seems likely that the cluster registry will need to also support delegated authn and authz, but that is not yet wired up. It should be straightforward implementation-wise, though whether this is just one of the authn/authz modes that can be enabled or a harder switch that makes the cluster registry API server flip from independent to dependent on another Kube API server is not yet resolved.

The cluster registry is currently only implemented as a standalone API server, with authn/authz fully independent of any hosting cluster. So, it could be run outside of a cluster, though the `crinit` tool bootstraps it into a cluster.

perotinus:
@pmorie @font I'm going to submit this later today if there are no objections.

PR #16, opened 2017-10-04
Flesh out the API of the cluster registry

perotinus:
The API is still a WIP, but this brings it closer to the expected finalform, and to what exists in the API design document.

cf https://docs.google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJlX1a960/edit# (Must be member of [kubernetes-dev](https://groups.google.com/forum/#!forum/kubernetes-dev) to view.)

### DIFF
--- a/pkg/apis/clusterregistry/types.go
+++ b/pkg/apis/clusterregistry/types.go
@@ -24,13 +24,84 @@ import (
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// Information about a cluster in a cluster registry.
+// Cluster contains information about a cluster in a cluster registry.
 type Cluster struct {
 	metav1.TypeMeta
 	// Standard object's metadata.
 	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
 	// +optional
 	metav1.ObjectMeta
+
+	// State is the state of the cluster. This is not a specification, and is not
+	// meant to be used by actively-reconciling controllers; it is also not
+	// a status, as it contains fields that do not necessarily describe the
+	// status of the cluster, and is not necessarily updated by an active
+	// controller.
+	// +optional
+	State ClusterState
+}
+
+// ClusterState contains the state of a cluster.
+type ClusterState struct {
+
+	// KubeAPIServer represents the API server for this cluster.
+	// +optional
+	KubeAPIServer KubeAPIServer
+
+	// AuthInfo contains public information that can be used to authenticate
+	// to and authorize with this cluster.
+	// +optional
+	AuthInfo AuthInfo
+
+	// CloudProvider contains information about the cloud provider this cluster
+	// is running on.
+	// +optional
+	CloudProvider CloudProvider
+}
+
+type URL string
+
+// KubeAPIServer represents one and only one Kubernetes API server.
+type KubeAPIServer struct {
+	// Servers specifies the addresses of the Kubernetes API server’s network
+	// identity or identities. They can be any valid HTTP URL including the
+	// IP:Port combination or the host name.
+	Servers []URL
+
+	// CABundle is the certificate authority information.
+	// +optional
+	CABundle []byte
+}
+
+// AuthInfo holds public information that describes how a client can get
+// credentials to access the cluster. For example, OAuth2 client registration
+// endpoints and supported flows, Kerberos servers locations, etc.
+//
+// It should not hold any private or sensitive information.
+type AuthInfo struct {
+
+	// AuthProviders is a list of configurations for auth providers.
+	AuthProviders []AuthProviderConfig
+}
+
+// AuthProviderConfig contains the information necessary for a client to
+// authenticate with a Kubernetes API server. It is modeled after
+// k8s.io/client-go/tools/clientcmd/api/v1.AuthProviderConfig.
+type AuthProviderConfig struct {
+	// Name is the name of this configuration.
+	Name string
+
+	// Config is a map of values that contains the information necessary for a
+	// client to authenticate to a Kubernetes API server.
+	// +optional
+	Config map[string]string
+}
+
+// CloudProvider contains information about the cloud provider this cluster is
+// running on.
+type CloudProvider struct {
+	// Name is the name of the cloud provider for this cluster.
+	Name string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/clusterregistry/v1alpha1/types.go
+++ b/pkg/apis/clusterregistry/v1alpha1/types.go
@@ -23,6 +23,7 @@ import (
 // +genclient
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
 // Information about a cluster in a cluster registry.
 type Cluster struct {
 	metav1.TypeMeta `json:",inline"`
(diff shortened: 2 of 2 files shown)


perotinus:
cc @font @madhusudancs

pkg/apis/clusterregistry/types.go
```diff
@@ -24,13 +24,71 @@ import (
...
-// Information about a cluster in a cluster registry.
+// Cluster contains information about a cluster in a cluster registry.
 type Cluster struct {
 	metav1.TypeMeta
 	// Standard object's metadata.
 	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
 	// +optional
 	metav1.ObjectMeta
+
+	// State is thestate of the cluster. This is not a specification, and is not
```
pmorie: typo: the state

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -30,6 +30,62 @@ type Cluster struct {
...
+	// a status, as it contains fields that do not necessarily describe the
+	// status of the cluster, and is not necessarily updated by an active
+	// controller.
+	State ClusterState
+}
+
+// ClusterState contains the state of a cluster.
+type ClusterState struct {
+	KubeAPIServer KubeAPIServer `json:"kube_api_server,omitempty" protobuf:"bytes,1,opt,name=kube_api_server"`
+	AuthInfo      AuthInfo      `json:"auth_info,omitempty" protobuf:"bytes,2,opt,name=auth_info"`
```
pmorie: If these fields are optional, they should be pointers if they do not have a meaningful zero type.  I think `KubeAPIServer` is not optional, `AuthInfo` is arguable not optional and `CloudProvider` should be optional.

WDYT?

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -30,6 +30,62 @@ type Cluster struct {
...
+
+// AuthInfo holds public information that describes how a client can get
+// credentials to access the cluster. For example, OAuth2 client registration
+// endpoints and supported flows, Kerberos servers locations, etc.
+//
+// It should not hold any private or sensitive information.
+type AuthInfo struct {
+
+	// AuthProviders is a list of configurations for auth providers.
+	AuthProviders []AuthProviderConfig `json:"auth_providers,omitempty" protobuf:"bytes,1,opt,name=auth_providers"`
```
pmorie: k8s convention is `authProviders` - is there a reason to do something differently?

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -30,6 +30,62 @@ type Cluster struct {
...
+}
+
+// AuthProviderConfig contains the information necessary for a client to
+// authenticate with a Kubernetes API server. It is modeled after
+// k8s.io/client-go/tools/clientcmd/api/v1.AuthProviderConfig.
+type AuthProviderConfig struct {
+	// Name is the name of this configuration.
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
pmorie: Required fields should not be `omitempty`.

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -30,6 +30,62 @@ type Cluster struct {
...
+
+	// Config is a map of values that contains the information necessary for a
+	// client to authenticate to a Kubernetes API server.
+	Config map[string]string `json:"config,omitempty" protobuf:"bytes,2,opt,name=config"`
```
pmorie: If this contains secret information, it will mean that `Cluster` is an escalating resource.  I don't think we should couple to k8s secrets, but we can probably factor differently and enable a view of `Cluster` that doesn't give you access to secret values in the Config.

pkg/apis/clusterregistry/types.go
perotinus: Thanks! Fixed.

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: No, no good reason at all. Fixed.

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: This should be fixed everywhere in the code now, though I haven't yet updated in response to your other comment.

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: I think the intention is that the secret values not be stored in Config at all. The cluster registry is not meant to hold secret values, but it is meant to hold information (e.g., webhook URI) that lets a user authenticate with the cluster in another manner.

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: That sounds reasonable to me, and I agree with your assessments about the optionality of the fields.

Do you think you could add a comment to [the API doc](https://docs.google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJlX1a960/edit#) in case people want to discuss it further there?

pkg/apis/clusterregistry/types.go
```diff
@@ -24,13 +24,84 @@ import (
...
+
+// AuthProviderConfig contains the information necessary for a client to
+// authenticate with a Kubernetes API server. It is modeled after
+// k8s.io/client-go/tools/clientcmd/api/v1.AuthProviderConfig.
+type AuthProviderConfig struct {
+	// Name is the name of this configuration.
+	Name string
+
+	// Config is a map of values that contains the information necessary for a
+	// client to authenticate to a Kubernetes API server.
```
pmorie: I think that if we do not want secret information to be stored in this resource, we should specifically call out like so:

```
Config should NOT contain secret information.
```

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -23,16 +23,89 @@ import (
...
+	// +optional
+	State ClusterState `json:"state,omitempty" protobuf:"bytes,1,opt,name=state"`
+}
+
+// ClusterState contains the state of a cluster.
+type ClusterState struct {
+
+	// KubeAPIServer represents the API server for this cluster.
+	// +optional
+	KubeAPIServer KubeAPIServer `json:"kubeApiServer,omitempty" protobuf:"bytes,1,opt,name=kubeApiServer"`
```
pmorie: I don't think this is optional :)

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -23,16 +23,89 @@ import (
...
 	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// State is the state of the cluster. This is not a specification, and is not
+	// meant to be used by actively-reconciling controllers; it is also not
+	// a status, as it contains fields that do not necessarily describe the
+	// status of the cluster, and is not necessarily updated by an active
+	// controller.
+	// +optional
+	State ClusterState `json:"state,omitempty" protobuf:"bytes,1,opt,name=state"`
```
pmorie: This shouldn't be optional

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -23,16 +23,89 @@ import (
...
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
 // Information about a cluster in a cluster registry.
 type Cluster struct {
 	metav1.TypeMeta `json:",inline"`
 	// Standard object's metadata.
 	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
 	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
pmorie: I don't think this should be optional

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: Agreed in principle, but see above comment. It looks like many (most? all?) objects mark their Spec and Status as optional, so it seems to make sense to do that here.

pkg/apis/clusterregistry/v1alpha1/types.go
perotinus: This makes conceptual sense to me, but it looks like many (most?) Kubernetes API objects do mark the ObjectMeta as optional.

https://github.com/kubernetes/api/blob/master/core/v1/types.go
https://github.com/kubernetes/api/blob/master/apps/v1beta2/types.go

I'm not sure I want to go against that convention.

pkg/apis/clusterregistry/v1alpha1/types.go
pmorie: bizarre and unexpected, but fine for now.

pkg/apis/clusterregistry/v1alpha1/types.go
pmorie: bizarre and unexpected, but fine for now

pmorie:
I'm fine with this going in - sorry for the reply latency.

perotinus:
Thank you, @pmorie! I'll open up an issue that tracks your concerns that haven't been explicitly addressed in this PR.

PR #17, opened 2017-10-05
Rename cmd/cr/app/clusterregistry.go to cluster.go

perotinus:
cc @madhusudancs @font


perotinus:
This was a mistake made in the large-scale renaming of things from federation to clusterregistry. This file installs the cluster API, so cluster.go is a more-appropriate name.

perotinus:
Good call, @font. Done!

PR #18, opened 2017-10-06
Update the codegen scripts to use the repos from the workspace rather than the local GOPATH

perotinus:
This is not a pretty solution, but until there is good enough support for generating code with genrules, this makes the building a bit easier and means that we're using the same versions as the rest of the code is being built against. This should also allow the cluster-registry repository to live anywhere, not just in GOPATH/src/k8s.io/cluster-registry

I attempted to test this with clean and dirty repositories, but if you have time to patch in and test, I would appreciate it. You may want to `bazel clean --expunge` first.

cc @madhusudancs @font


perotinus:
Note that this currently has an issue with generated defaults and deepcopies which I'm looking into.

perotinus:
OK, this should work for all generated files now.

hack/update-codegen.sh
```diff
@@ -18,25 +18,29 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_PACKAGE=k8s.io/cluster-registry
-SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")
+SCRIPT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
madhusudancs: Isn't this equivalent to just running `"$( dirname "${BASH_SOURCE[0]}" )"`? Is this to expand relative path to absolute path?

hack/update-codegen.sh
```diff
@@ -18,25 +18,29 @@ set -o errexit
...
 SCRIPT_BASE=$(cd ${SCRIPT_ROOT}/../..; pwd)
+SCRIPT_PACKAGE=$(basename $(dirname "${SCRIPT_ROOT}"))
```
madhusudancs: This only works if the script is only one level below the real source root right? Why not just encode that here instead of this indirect encoding?

hack/update-codegen.sh
```diff
@@ -50,14 +54,28 @@ function generate_group() {
...
+  if [[ "${GEN_TMPDIR}" == "/tmp/gen"* ]]; then
+    rm -rf "${GEN_TMPDIR}"
+  fi
 }
 
 generate_group clusterregistry v1alpha1
+
+if [[ "${TMP_GOPATH}" == "/tmp/gopath"* ]]; then
+  rm -rf "${TMP_GOPATH}"
+fi
```
madhusudancs: Do you want to make this a trap handler? See https://github.com/kubernetes/kubernetes/blob/04477ff90ceb55b864a4ac96fcc8c8b510414ead/staging/src/k8s.io/apiextensions-apiserver/hack/build-image.sh#L25 for an example.

hack/gopath_from_workspace.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+#
```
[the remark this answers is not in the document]
perotinus: Good point. Done.

hack/update-codegen.sh
perotinus: I'm not quite sure I follow what you're asking me to do. What would be the direct version of encoding this?

hack/update-codegen.sh
perotinus: Good call. Done.

hack/update-codegen.sh
perotinus: Yes, that's the idea. Under the right circumstances, they should be equivalent, but I prefer to make sure the path is an absolute path.

hack/update-codegen.sh
```diff
@@ -50,14 +58,28 @@ function generate_group() {
...
+  rsync -a "${GEN_TMPDIR}/k8s.io/cluster-registry/" "${SCRIPT_BASE}/${SCRIPT_PACKAGE}"
+
+  if [[ "${GEN_TMPDIR}" == "/tmp/gen"* ]]; then
+    rm -rf "${GEN_TMPDIR}"
+  fi
 }
 
+echo "Creating temporary GOPATH from bazel workspace"
+TMP_GOPATH="$("${SCRIPT_ROOT}/gopath_from_workspace.sh")"
+trap cleanup EXIT
```
ericchiang: s/cleanup/clean_up/g

Hit this running locally:

```
$ cd $( mktemp -d )
$ GOPATH=$PWD
$ git clone https://github.com/perotinus/cluster-registry.git src/k8s.io/cluster-registry
$ cd src/k8s.io/cluster-registry
$ git checkout useworkspacereposforgen
$ bazel clean --expunge
$ ./hack/update-codegen.sh
# lot of output
./hack/update-codegen.sh: line 1: cleanup: command not found   
```

ericchiang:
Is this exercised in CI?

hack/update-codegen.sh
perotinus: Thanks, fixed!

perotinus:
It is not exercised in CI; we haven't set up CI for this repository yet. It's on my todo list.

perotinus:
@font @madhusudancs @ericchiang I'm planning to submit this today if there are no additional comments.

hack/update-codegen.sh
madhusudancs: Ok. This is usually achieved through `readlink -f`, but that also follows symlinks. What you have now works too.

hack/update-codegen.sh
madhusudancs: By direct encoding I meant specifying that you are asking for the parent's basename by writing something like `basename $(readlink -f "${SCRIPT_ROOT}/..")`. But this is fine.

hack/update-codegen.sh
```diff
@@ -18,25 +18,33 @@ set -o errexit
...
 SCRIPT_BASE=$(cd ${SCRIPT_ROOT}/../..; pwd)
+SCRIPT_PACKAGE=$(basename $(dirname "${SCRIPT_ROOT}"))
 
+# Called on EXIT after the temporary directory is created.
+function clean_up() {
+  if [[ "${TMP_GOPATH}" == "/tmp/gopath"* ]]; then
+    rm -rf "${TMP_GOPATH}"
+  fi
+}
+
```
madhusudancs: It's more readable if you move line 83 `trap clean_up EXIT` here.

hack/gopath_from_workspace.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/sh
+
+# Copyright 2014 The Kubernetes Authors.
```
madhusudancs: s/2014/2017/

perotinus: Thanks!

hack/update-codegen.sh
perotinus: Done.

perotinus:
@madhusudancs PTAL, I updated the script structure a bit to make the temporary directory deletion cleaner.

Issue #20, opened 2017-10-10
Cluster API issues


perotinus:
@pmorie raised a few issues in #16 about the cluster registry API that were not addressed:

- [x] being more explicit about how credentials are handled/whether credentials should be stored
  - work here is waiting for input from SIG-auth about how to handle cluster credentials in the cluster registry
- [x] optional fields should be pointers
  - this seems reasonable, but would like to discuss with sig-arch and/or sig-api-machinery to make sure

Besides this, there are a few issues from the cluster registry API design that also need to be addressed:
- [x] using a stronger type for the Kubernetes API server info
  - this was going to be a copy of `api.Endpoints` from k8s, for reasons described [here](https://docs.google.com/a/google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJlX1a960/edit?disco=AAAABFxkK0A), but after pushback more discussion is needed
- [x] spec/status field names/presence
  - this is waiting on discussions with sig-arch and/or sig-api-machinery
- [x] `CABundle` in `KubeAPIServer`
  - waiting for response on comment

And an issue from #29:
- [x] figure out motivation for or remove `CloudProvider` field

cf https://docs.google.com/a/google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJlX1a960/edit?disco=AAAABFxkK0A

cc @pmorie @quinton-hoole

perotinus:
cc @madhusudancs

perotinus:
Updates for the above points:

1. After discussion with SIG-auth, we made the decision to go with a generic type for auth, and to heavily discourage storing credentials in the cluster registry
3. We will use an Endpoints object copied from the k8s API.
4. We resolved these discussions in favor of spec/status
5. Assuming lazy consensus, we are going with what's there now.

madhusudancs:
I made this comment on the PR #31. But I am noticing this issue now. So copying the comment here. 

I am really confused. Copying Endpoints object seems incorrect. For example, what purpose would EndpointPort.Protocol serve? If anything, this should mirror what's in federation now. I can't see all the fields in Kubernetes core api/v1.Endpoint being appropriate here.

madhusudancs:
I think if we are copying anything, what's in federation today is far more useful than copying `api/v1.Endpoint`.

perotinus:
I believe that all of these issues have been addressed. Closing.

PR #31, opened 2017-10-17
Update the cluster API per the latest decisions

perotinus:
This is intended to get the API very close to an alpha state.

cf #20 

cc @font @madhusudancs @pmorie

### DIFF
--- a/pkg/apis/clusterregistry/types.go
+++ b/pkg/apis/clusterregistry/types.go
@@ -32,67 +32,90 @@ type Cluster struct {
 	// +optional
 	metav1.ObjectMeta
 
-	// State is the state of the cluster. This is not a specification, and is not
-	// meant to be used by actively-reconciling controllers; it is also not
-	// a status, as it contains fields that do not necessarily describe the
-	// status of the cluster, and is not necessarily updated by an active
-	// controller.
+	// Spec is the specification of the cluster. This may or may not be reconciled
+	// by an active controller.
 	// +optional
-	State ClusterState
+	Spec ClusterSpec
+
+	// Status is the status of the cluster. It is optional, and can be left nil
+	// to imply that the cluster status is not being reported.
+	// +optional
+	Status *ClusterStatus
 }
 
-// ClusterState contains the state of a cluster.
-type ClusterState struct {
-
-	// KubeAPIServer represents the API server for this cluster.
+// ClusterSpec contains the spec of a cluster.
+type ClusterSpec struct {
+	// KubernetesAPIEndpoints represents the endpoints of the API server for this
+	// cluster.
 	// +optional
-	KubeAPIServer KubeAPIServer
+	KubernetesAPIEndpoints KubernetesAPIEndpoints
 
 	// AuthInfo contains public information that can be used to authenticate
-	// to and authorize with this cluster.
+	// to and authorize with this cluster. It is not meant to store private
+	// information (e.g., tokens or client certificates) and cluster registry
+	// implementations are not expected to provide hardened storage for
+	// secrets.
 	// +optional
 	AuthInfo AuthInfo
 
 	// CloudProvider contains information about the cloud provider this cluster
 	// is running on.
 	// +optional
-	CloudProvider CloudProvider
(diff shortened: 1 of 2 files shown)


madhusudancs:
I am really confused. Copying Endpoints object seems incorrect. For example, what purpose would `EndpointPort.Protocol` serve? If anything, this should mirror what's in federation now. I can't see all the fields in Kubernetes core api/v1.Endpoint being appropriate here.

perotinus:
@madhusudancs If the concern is that `EndpontPort.Protocol` is not useful, we can remove it.

@quinton-hoole, what was your reasoning behind suggesting using`api.Endpoints` rather than mimicking what Federation currently has?

perotinus:
@madhusudancs, are you OK with this going in like this for now, with the linked issue preserving a question about using the Endpoints object?

madhusudancs:
I am Ok with this change going in without the Endpoints change. It's more work to revert it than separating the parts under question into a separate commit.

perotinus:
@font @pmorie @madhusudancs Any other issues? I'd like to submit this; we can iterate further if necessary.

madhusudancs:
@perotinus I don't see any change to the PR since I last looked. Did you push a new commit? I am still on removing the endpoints change in this PR until it is settled. No other issues. If you remove that change along with the dependent types, I am fine lgtm'ing this PR.

tengqm:
I tend to agree with @madhusudancs that introducing a new 'Endpoint' concept may cause confusion to people. We have to explain to users that this Endpoint is not the one they see from kubernetes ... If possible, we may want to avoid the Cartesian product computation ... we may want to support just HTTP/HTTPS (instead of UDP/TCP)... After all, we are talking about a Kubernetes API server instead of a generic service endpoint to be supported by Kubernetes.

ericchiang:
We use a load balancer behind a DNS address for our HA API servers. All clients use the load balancer address. We never refer to the API servers through a bare IP address.

Why did the API switch to only supporting IPs?

perotinus:
@ericchiang The API as it is still supports hostnames: https://github.com/kubernetes/cluster-registry/pull/31/files#diff-7d367230bef0cc50523964c6d3dccdcbR206 

I admit the naming is confusing.

Given the amount of pushback, I'm going to change this to something like what Federation does today. I believed that this had been addressed in the API doc, since there was a lack of objection to using `api.Endpoints`, but we'll have to have some more explicit discussion in a follow-up.

cc @quinton-hoole, who suggested the `api.Endpoints` type

perotinus:
@ericchiang @madhusudancs PTAL.

pmorie:
Soery i missed your earlier tag - gh notifications are pretty useless for
me. Will TALat this tonight.

pkg/apis/clusterregistry/types.go
```diff
@@ -32,43 +32,64 @@ type Cluster struct {
...
+}
+
+// ClusterStatus contains the status of a cluster.
+type ClusterStatus struct {
+	// TODO https://github.com/kubernetes/cluster-registry/issues/28
 }
 
-type URL string
+// ServerAddressByClientCIDR helps the client to determine the server address that they should use, depending on the clientCIDR that they match.
+type ServerAddressByClientCIDR struct {
```
font: small nit: can we define this type below the `KubernetesAPIEndpoints` type?

perotinus:
/assign @pmorie 

@pmorie No worries; would it help if I assign things to you directly?

pkg/apis/clusterregistry/types.go
perotinus: Sure, done.

pmorie:
@perotinus unfortunately I think i need to rejigger my filters to have a hope of making notifications on github useful again.  I actually watch this repo and have every go into my inbox, but it's a flood.

pmorie:
LGTM, i think i may open a PR with a couple tweeks but this is movement in the right direction.

madhusudancs:
/lgtm

Thanks for modifying the endpoint object.

perotinus:
Thanks, everyone!

PR #32, opened 2017-10-17
Fix the crinit command descriptions.

perotinus:


perotinus:
cc @madhusudancs @font

PR #34, opened 2017-10-17
 Add a basic integration test for the cluster registry

perotinus:
This is based on work done for the kube-apiserver in [k/k/46865](https://github.com/kubernetes/kubernetes/pull/46865).

### DIFF
--- a/pkg/clusterregistry/testing/server_test.go
+++ b/pkg/clusterregistry/testing/server_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testing
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cluster-registry/pkg/apis/clusterregistry/v1alpha1"
+	crclientset "k8s.io/cluster-registry/pkg/client/clientset_generated/clientset"
+)
+
+func TestCreateAndGet(t *testing.T) {
+	config, tearDown := StartTestServerOrDie(t)
+	defer tearDown()
+
+	clientset, err := crclientset.NewForConfig(config)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	_, err = clientset.ClusterregistryV1alpha1().Clusters().Create(&v1alpha1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+	})
+
+	cluster, err := clientset.ClusterregistryV1alpha1().Clusters().Get("cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if cluster == nil {
+		t.Fatalf("Expected a cluster, got nil")
+	}
+	if cluster.Name != "cluster" {
+		t.Fatalf("Expected a cluster named 'cluster', got a cluster named '%v.", cluster.Name)
+	}
+}
--- a/pkg/clusterregistry/testing/testserver.go
+++ b/pkg/clusterregistry/testing/testserver.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testing
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	etcdtesting "k8s.io/apiserver/pkg/storage/etcd/testing"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/cluster-registry/pkg/apis/clusterregistry/install"
+	"k8s.io/cluster-registry/pkg/clusterregistry"
+	"k8s.io/cluster-registry/pkg/clusterregistry/options"
+)
+
+// TearDownFunc is to be called to tear down a test server.
+type TearDownFunc func()
+
+// StartTestServerOrDie calls startTestServer with up to 5 retries on bind error
+// and dies with t.Fatal if it does not succeed.
+func StartTestServerOrDie(t *testing.T) (*restclient.Config, TearDownFunc) {
+	// Retry starting the server because the bind might fail due to a race with
+	// another process binding to the port. We cannot listen to :0 (then the
+	// kernel would give us a port which is free for sure), so we need this
+	// workaround.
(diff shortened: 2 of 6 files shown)


perotinus:
cc @font @madhusudancs @marun

pmorie:
@perotinus seems like a bot is asleep at the wheel

perotinus:
@pmorie Yeah, it seems so. I'm going to push a new PR and see if it finds that one.

perotinus:
@marun @madhusudancs @font Friendly ping?

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, nil, fmt.Errorf("Failed to wait for /healthz to return OK: %v", err)
+	}
+
+	return server.LoopbackClientConfig, tearDown, nil
+}
+
+// freePort finds and returns a free port on the local machine. Will return -1
+// for the port if there is an error.
+func freePort() (int, error) {
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
```
madhusudancs: This doesn't seem consistent with the comment in line 44 - https://github.com/kubernetes/cluster-registry/pull/34/files#diff-096c0acbe4241fc9010c8fd6e54c914bR44. Am I missing something?

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return -1, err
+	}
+
+	l, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		return -1, err
+	}
+	defer l.Close()
```
madhusudancs: Also, why open and close here? Is it not sufficient to just return `addr.Port`?

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	runErrCh := make(chan error, 1)
+	go func(stopCh <-chan struct{}) {
+		if err := server.PrepareRun().Run(stopCh); err != nil {
+			t.Logf("cluster registry exited uncleanly: %v", err)
+			runErrCh <- err
+		}
+	}(stopCh)
+
+	t.Logf("Waiting for /healthz to be OK...")
```
madhusudancs: I would log this after client construction and log "Creating a new client for cluster registry API server" or something like that here.

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+		select {
+		case err := <-runErrCh:
+			return false, err
+		default:
+		}
+
+		result := client.CoreV1().RESTClient().Get().AbsPath("/healthz").Do()
+		status := 0
+		result.StatusCode(&status)
+		if status == 200 {
```
madhusudancs: Use `http.StatusOK` instead of 200 - https://golang.org/pkg/net/http/#pkg-constants.

You could write this as `return status == http.StatusOK, nil` and remove the next 3 lines.

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to wait for /healthz to return OK: %v", err)
```
madhusudancs: I am not entirely sure if this error message is appropriate. It should be something like - "failed waiting for /healthz to ...", no?

pkg/clusterregistry/testing/server_test.go
```diff
@@ -0,0 +1,49 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+	})
+
+	cluster, err := clientset.ClusterregistryV1alpha1().Clusters().Get("cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if cluster == nil {
```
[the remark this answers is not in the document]
perotinus: Hmm. I'm not sure what additional useful testing I can do here: I created an empty cluster named "cluster" and retrieved a cluster named "cluster". What did you have in mind?

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+	// another process binding to the port. We cannot listen to :0 (then the
+	// kernel would give us a port which is free for sure), so we need this
+	// workaround.
+	for retry := 0; retry < 5 && !t.Failed(); retry++ {
+		config, td, err := startTestServer(t)
+		if err == nil {
+			return config, td
+		}
+		if err != nil && !strings.Contains(err.Error(), "bind") {
+			break
```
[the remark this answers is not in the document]
perotinus: That makes sense. Replaced `break` with a `t.Fatalf` call.

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		t.Logf("Bind error, retrying...")
+	}
+
+	t.Fatalf("Failed to launch server")
+	return nil, nil
+}
+
+// startTestServer starts a etcd server and cluster registry API server. Returns
```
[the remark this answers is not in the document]
perotinus: Done.

pkg/clusterregistry/testing/testserver.go
perotinus: Done.

pkg/clusterregistry/testing/testserver.go
```diff
@@ -0,0 +1,159 @@
...
+// a REST client config and a tear-down function.
+//
+// Note: we return a tear-down func instead of a stop channel because the latter
+//       will leak temporary files that becaues Golang testing's call to os.Exit
```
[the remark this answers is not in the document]
perotinus: Done.

pkg/clusterregistry/testing/testserver.go
perotinus: Oh, I like the condensing. Done.

pkg/clusterregistry/testing/testserver.go
perotinus: The comment in line 44 is talking about `options.SecureServing.BindPort`, whose semantics are different: for that field, if it is set to 0, then HTTPS serving is disabled rather than the system attempting to find a port. Setting the port to 0 as a way to find an open port is for the system libraries, which is why it's being done here.

pkg/clusterregistry/testing/testserver.go
perotinus: Good point. I've updated it.

pkg/clusterregistry/testing/testserver.go
perotinus: I'm not exactly sure. I'm assuming that the way to check that the port is free is to actually attempt to listen.

I've simplified this a bit, though the logic should be conceptually the same.

pkg/clusterregistry/testing/server_test.go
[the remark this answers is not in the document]
madhusudancs: The cluster name? As far as I can tell, you are asking for a particular cluster, but you are not checking if that's what is being returned.

pkg/clusterregistry/testing/testserver.go
madhusudancs: I see Ok. Didn't realize that.

pkg/clusterregistry/testing/testserver.go
madhusudancs: I think binding it shouldn't be necessary, but this is good for now.

pkg/clusterregistry/testing/server_test.go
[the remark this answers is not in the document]
perotinus: Good point, I've added this check.

perotinus:
@madhusudancs @font Can you take a look? I had to make some code changes to fix an issue caught by this test following the vendoring update.

pkg/clusterregistry/testing/testserver.go
```diff
@@ -49,7 +50,8 @@ func StartTestServerOrDie(t *testing.T) (*restclient.Config, TearDownFunc) {
 		if err == nil {
 			return config, td
 		}
-		if err != nil && !strings.Contains(err.Error(), "bind") {
+		if !strings.Contains(err.Error(), "bind") {
+			t.Fatalf("Failed to launch server: %v", err)
 			break
```
madhusudancs: Not entirely sure if `t.Fatalf` replaced `break`. I still see `break` here. But you can fix this in a follow up PR.

perotinus: Whoops, that was not intentional! It's easy enough to fix up here, so I'll do that.

madhusudancs: Thanks for fixing it.

PR #40, opened 2017-10-18
Simplify the crinit command scaffolding a bit further

perotinus:


perotinus:
cc @font @madhusudancs

PR #44, opened 2017-10-20
Fix a few issues identified by go vet

perotinus:
cc @font @madhusudancs


perotinus:
@font No, I ran this locally out of curiosity and found these issues. I'd like to get `go vet` running as a presubmit check, but I wanted to get these in before they got lost.

PR #45, opened 2017-10-20
Update the codegen and test scripts to work on Prow

perotinus:
cf https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/cluster-registry/44/pull-cluster-registry-bazel/3/


perotinus:
I'm looking into working around the `BASH_SOURCE` issue now. Oddly, it seems like the issue is only in the child script, not `update-codegen.sh`, which doesn't quite make sense to me, but leaves a potential straightforward workaround strategy available.

perotinus:
OK, this should be ready for review now.

I'm going to work on getting the codegen scripts to run directly from bazel, so that the chained scenario calls will not be required and this should become simpler, but this works for now.

perotinus:
@madhusudancs @font Ping?

Issue #24, opened 2017-10-13
Determine whether generated code needs to be checked in


perotinus:
From a comment on #21, there might be reason to check in the generated artifacts to enable people who do not use bazel to depend on this repository; and perhaps to make the test infrastructure easier to wrangle. I am quite opposed to checking in artifacts unless it really is blocking a workflow, so this issue is tracking whether enough need to outweigh the pain of checked-in generated code.

If this is causing you problems, please bump this issue!

pmorie:
I have a strong preference toward:

- Checking in any generated code
- Checking in vendored dependencies

ericchiang:
Regardless of checking in vendored code, I would like to stress that cluster-registry must check in generated code to be compatible with non-bazel based project.

As someone who needs to use the cluster-registry client and coming from an org that hasn't used bazel in the past, this is a blocker.

pmorie:
Yes, I would prefer to check in any generated code as well.

ericchiang:
Another note: because this repo doesn't commit its BUILD.bazel files, it's not possible to depend on the generated code even if you're using bazel. (I might be wrong since I'm newer to bazel)

perotinus:
Generated code and BUILD files are now checked in.

PR #53, opened 2017-10-25
*: move docker build to top level BUILD.bazel file

ericchiang:
Right now, executing `bazel run //:gazelle` results in a dirty git repo, since `/cmd/clusterregistry/BUILD.bazel` was previously ignored. Move the docker build logic to the root  BUILD.bazel file, so all generated code is properly ignored.

cc @perotinus any thoughts here? It's kinda weird that building the binary results in a dirty git repo.


ericchiang:
Nice to see CI's working. :)

I'm going to guess that it'll be easier to just solve https://github.com/kubernetes/cluster-registry/issues/24 than mess with this.

perotinus:
@ericchiang Yes, that's possibly true. There seems to be demand for it from two different integrators, so I think in balance I'll have to forego my crusade to avoid checked-in generated code and wait for k/k to blaze that trail.

That said, if you want to fix this, you need to modify [this](https://github.com/kubernetes/cluster-registry/blob/master/hack/bazel-test.sh#L32) and change the docker_push target to reference the new location.

perotinus:
Now that #24 is solved, this change is no longer necessary.

/close

PR #57, opened 2017-10-30
Update the cluster-registry dependencies to newer versions.

perotinus:
This is a first step towards checked-in vendored dependencies.

- Update the Go rules to use the same version as used by `k/k`
- Update the `k/` dependencies to post-1.8 versions.

The versions used here were generally those pulled by the `dep` tool, which I have been using for vendoring. A future PR will replace the `WORKSPACE` dependencies with a vendor tree.

/assign @madhusudancs, @pmorie 
cc @font

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,12 +1,37 @@
-git_repository(
+# These dependencies' versions are pulled from the k/k WORKSPACE.
+# https://github.com/kubernetes/kubernetes/blob/71624d85fab78a1b9d63f7def1e500ad48806620/build/root/WORKSPACE
+http_archive(
     name = "io_bazel_rules_go",
-    remote = "https://github.com/bazelbuild/rules_go.git",
-    tag = "0.5.5",
+    sha256 = "441e560e947d8011f064bd7348d86940d6b6131ae7d7c4425a538e8d9f884274",
+    strip_prefix = "rules_go-c72631a220406c4fae276861ee286aaec82c5af2",
+    urls = ["https://github.com/bazelbuild/rules_go/archive/c72631a220406c4fae276861ee286aaec82c5af2.tar.gz"],
 )
 
-load("@io_bazel_rules_go//go:def.bzl", "go_repositories", "go_repository")
+http_archive(
+    name = "io_kubernetes_build",
+    sha256 = "8e49ac066fbaadd475bd63762caa90f81cd1880eba4cc25faa93355ef5fa2739",
+    strip_prefix = "repo-infra-e26fc85d14a1d3dc25569831acc06919673c545a",
+    urls = ["https://github.com/kubernetes/repo-infra/archive/e26fc85d14a1d3dc25569831acc06919673c545a.tar.gz"],
+)
 
-go_repositories()
+load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains", "go_repository")
+
+go_rules_dependencies()
+
+go_register_toolchains(go_version = "1.9.1")
+
+# This has to go before proto_register_toolchains() or else that will pull in
+# an incompatible version.
+go_repository(
+    name = "org_golang_google_grpc",
+    build_file_proto_mode = "disable",
+    commit = "bfaf0423469fc4f95e9eac7e3eec0b2abb46fcca",
+    importpath = "google.golang.org/grpc",
+)
+
+load("@io_bazel_rules_go//proto:def.bzl", "proto_register_toolchains")
+
+proto_register_toolchains()
 
 # Docker rules
 git_repository(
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -1,12 +1,35 @@
-git_repository(
+http_archive(
     name = "io_bazel_rules_go",
-    remote = "https://github.com/bazelbuild/rules_go.git",
-    tag = "0.5.5",
+    sha256 = "441e560e947d8011f064bd7348d86940d6b6131ae7d7c4425a538e8d9f884274",
+    strip_prefix = "rules_go-c72631a220406c4fae276861ee286aaec82c5af2",
+    urls = ["https://github.com/bazelbuild/rules_go/archive/c72631a220406c4fae276861ee286aaec82c5af2.tar.gz"],
```
madhusudancs: Do they also have a corresponding release version or a tag along with the SHA? If there is a human-readable equivalent for this SHA, could you please add that as a comment? Because now, I don't know why this revision was used instead of a release and how to choose a new revision in the future.

WORKSPACE
```diff
@@ -1,12 +1,35 @@
...
 )
 
-load("@io_bazel_rules_go//go:def.bzl", "go_repositories", "go_repository")
+http_archive(
+    name = "io_kubernetes_build",
+    sha256 = "8e49ac066fbaadd475bd63762caa90f81cd1880eba4cc25faa93355ef5fa2739",
+    strip_prefix = "repo-infra-e26fc85d14a1d3dc25569831acc06919673c545a",
+    urls = ["https://github.com/kubernetes/repo-infra/archive/e26fc85d14a1d3dc25569831acc06919673c545a.tar.gz"],
```
madhusudancs: Same here.

WORKSPACE
```diff
@@ -1,12 +1,35 @@
...
+go_rules_dependencies()
+
+go_register_toolchains(go_version = "1.9.1")
+
+# This has to go before proto_register_toolchains() or else that will pull in
+# an incompatible version.
+go_repository(
+    name = "org_golang_google_grpc",
+    build_file_proto_mode = "disable",
+    commit = "bfaf0423469fc4f95e9eac7e3eec0b2abb46fcca",
```
madhusudancs: Same here.

WORKSPACE
```diff
@@ -40,29 +63,32 @@ go_repository(
     name = "io_k8s_apimachinery",
     build_file_generation = "on",
     build_file_name = "BUILD.bazel",
-    commit = "b166f81f5c4c88402ae23a0d0944c6ad08bffd3b",
+    build_file_proto_mode = "disable",
+    commit = "18a564baac720819100827c16fdebcadb05b2d0d",
```
madhusudancs: I guess these SHAs (this and the subsequent ones) were all just matched to the ones in k/k? Can that be a single comment above this rule for all the dependencies that explains how the SHAs were chosen ?

pkg/clusterregistry/server.go
```diff
@@ -57,7 +57,7 @@ func NonBlockingRun(s *options.ServerRunOptions, stopCh <-chan struct{}) error {
 	if err := s.SecureServing.MaybeDefaultWithSelfSignedCerts(s.GenericServerRunOptions.AdvertiseAddress.String(), nil, nil); err != nil {
 		return fmt.Errorf("error creating self-signed certificates: %v", err)
 	}
-	s.SecureServing.ForceLoopbackConfigUsage()
+	//s.SecureServing.ForceLoopbackConfigUsage()
```
madhusudancs: Why?

perotinus: This method was removed from `SecureServing`. I forgot to remove the commented-out code!

WORKSPACE
perotinus: These were based on the result of running `dep` and seeing which.

k/ doesn't have a good solution for vendored dependency version management at this point. I don't want to solve that problem right now for this repository, and I think it's important to get vendoring in soon in order to unblock development.

WORKSPACE
perotinus: See above.

WORKSPACE
perotinus: See above.

WORKSPACE
perotinus: I don't believe so. I pulled these from k/k, and have added a comment to that effect.

perotinus:
Thanks, PTAL!

I think there is more work to do in the future with getting vendoring and dependency version management working in a consistent manner, but this is a larger k8s problem that other people are working on, so I'd like to get this update (and following updates) in to unblock other development.

WORKSPACE
madhusudancs: Top-level comment helps a lot. Thanks for adding it!

WORKSPACE
madhusudancs: I wasn't expecting you to solve the overall k8s vendoring problem and I understand getting this in is critical to unblock development. I was just expecting a comment in this file (or probably in a contributor doc or somewhere) explaining how these SHAs were generated and how could one update them in the future: links to files to copy from, commands to run, etc. Because before this comment, the PR looked like a black box. The top-level comment helps, if that is all one has to do to update all these SHAs.

perotinus: OK, that makes sense. I've added a brief comment here as well to explain how these were selected, but since this is quickly going to be replaced by #58, I didn't flesh out the comment in full detail.

PR #58, opened 2017-10-31
 Use dep to vendor dependencies into the repo

perotinus:
The commit "`Add vendored dependencies into the tree.`", is a generated comment. All hand-made changes are in the other commits.


ericchiang:
@perotinus per https://github.com/kubernetes/client-go/pull/319 because dep doesn't parse transitive Godeps.json files, if we want to use dep to vendor Kubernetes code, we need to specify _all_ of the packages in Gopkg.toml. Based on the generated Gopkg.lock, it looks like this PR is missing a few.

perotinus:
@ericchiang Hmm. Is that actually the case? I'm not sure that this needs to specify all of the dependencies as long as it is getting the correct ones. With the `Gopkg.lock` checked in, the dependency versions are fixed and recorded appropriately, so as long as they are working properly at this point, then we can defer adding more dependencies until it is necessary to update the vendored dependencies.

Updating vendored dependencies is not necessarily a common operation, and is quickly evolving, so I'd rather not spend time now working on a solution that targets today's issues if those may have changed by the next time it's necessary to update dependencies.

ericchiang:
> Updating vendored dependencies is not necessarily a common operation, and is quickly evolving, so I'd rather not spend time now working on a solution that targets today's issues if those may have changed by the next time it's necessary to update dependencies.

A TODO works for me.

/test pull-cluster-registry-bazel

ericchiang:
Hmmm, am I not allowed to re-run the tests?

perotinus:
@ericchiang That's odd, you should be able to if you're in the k8s org, I thought.

FWIW, I'm fixing an issue in the tests right now, so they will fail anyway.

perotinus:
OK, I've made some changes:
- fixes to (hopefully) make the tests work
- pruning unnecessary dependencies, which fell out of the above, since etcd has a recursive symlink in its `cmd` directory that bazel cannot handle

Gopkg.toml
```diff
@@ -0,0 +1,71 @@
...
+  # For verify/update scripts
+  "k8s.io/code-generator/cmd/client-gen",
+  # For verify scripts.
+  "github.com/kubernetes/repo-infra/verify/boilerplate/test",
+]
+
+ignored = ["github.com/kubernetes/repo-infra/kazel"]
+
+[[constraint]]
+  branch = "master"
```
madhusudancs: Is this is the end state and the plan? Or are you planning to pin this and other dependencies to specific versions? Seems like a lot of room for flakes.

docs/development.md
```diff
@@ -11,23 +24,21 @@ NOTE: There is an issue with bazel 0.6.x. As a workaround, use 0.5.x, or pass
...
+1.  Run `docker-credential-gcr configure-docker`
 
-Before doing any development work, you must (in order, from the repository root
-directory, after cloning):
+### dep
 
-1.  run `./hack/update-codegen.sh`
-1.  run `bazel run //:gazelle`
+This repository maintains its `vendor` directory with
+[dep](https://github.com/golang/dep). You must have a recent version of this
```
madhusudancs: Can you say how recent?

docs/development.md
```diff
@@ -72,9 +72,22 @@ from the repository root. (This may take a while the first time you run it.)
...
+    vendored dependencies or the `Gopkg.toml` file. The known ones are noted
+    below. Make these and any additional necessary ones, and add them to this
+    list.
+    -   As of #58, it was necessary to modify the BUILD file in
+        `vendor/k8s.io/client-go/util/cert` to have the go_library not reference
+        testdata.
 1.  Run the commands from the [Prerequisites](#Prerequisites).
 1.  Run the tests and fix any breakages.
+1.  When sending out a PR, please put the handmade changes in one commit and the
+    mechanical updates in another commit so that it's easier for reviewers to
```
madhusudancs: I think it will be more straightforward to say generated updates instead of mechanical updates.

Gopkg.toml
perotinus: The end state is still not quite well-defined. IMO, I think having this file constrain to master is reasonable, since the dependencies are checked in and `Gopkg.lock` has them precisely defined. There is an issue (#59) tracking doing a more-detailed strategy for vendoring.

docs/development.md
perotinus: Done.

docs/development.md
perotinus: Done.

Gopkg.toml
madhusudancs: Ok, fair enough.

Gopkg.toml
```diff
@@ -0,0 +1,71 @@
...
+[[constraint]]
+  name = "github.com/spf13/pflag"
+  version = "1.0.0"
+
+[[constraint]]
+  branch = "master"
+  name = "k8s.io/api"
+
+[[constraint]]
+  branch = "master"
```
pmorie: I would definitely not pin to master of this repo, especially since you have pinned to a specific hash of apiserver.  Likewise with client-go and any other k8s repo.  We will be constantly broken unless we pin to a known set of versions that works well together.

So, I would ensure that we are pinned to known versions of all k8s repos before merging this PR.

perotinus: That makes sense. I had left this like this because the specific version pinning is done in `Gopkg.lock`, and this file will need to be updated regardless when the tool is run again in order to get specific versions at that point.

However, I've updated this accordingly, since I agree that it looks confusing to have `master` in `Gopkg.toml`, and that specific versions maps better to how we actually have to do dependency management for now.

pmorie:
LGTM from me

PR #60, opened 2017-11-01
Tweak the setup of the crinit command to improve its usability

perotinus:
@font @pmorie @ericchiang @madhusudancs

### DIFF
--- a/pkg/crinit/crinit.go
+++ b/pkg/crinit/crinit.go
@@ -17,32 +17,35 @@ limitations under the License.
 package crinit
 
 import (
+	"flag"
 	"io"
 
-	"k8s.io/apiserver/pkg/util/flag"
+	apiserverflag "k8s.io/apiserver/pkg/util/flag"
 	"k8s.io/client-go/tools/clientcmd"
 	crinitinit "k8s.io/cluster-registry/pkg/crinit/init"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 // NewClusterregistryCommand creates the `clusterregistry` command.
 func NewClusterregistryCommand(out io.Writer, defaultServerImage, defaultEtcdImage string) *cobra.Command {
-	cmds := &cobra.Command{
+	rootCmd := &cobra.Command{
 		Use:   "crinit",
 		Short: "crinit runs a cluster registry in a Kubernetes cluster",
 		Long:  "crinit bootstraps and runs a cluster registry as a Deployment in an existing Kubernetes cluster.",
-		Run:   runHelp,
 	}
 
-	// From this point and forward we get warnings on flags that contain "_" separators
-	cmds.SetGlobalNormalizationFunc(flag.WarnWordSepNormalizeFunc)
+	// Add the command line flags from other dependencies (e.g., glog), but do not
+	// warn if they contain underscores.
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.CommandLine.SetNormalizeFunc(apiserverflag.WordSepNormalizeFunc)
+	rootCmd.PersistentFlags().AddFlagSet(pflag.CommandLine)
 
-	cmds.AddCommand(crinitinit.NewCmdInit(out, clientcmd.NewDefaultPathOptions(), defaultServerImage, defaultEtcdImage))
+	// Warn for other flags that contain underscores.
+	rootCmd.SetGlobalNormalizationFunc(apiserverflag.WarnWordSepNormalizeFunc)
 
-	return cmds
-}
+	rootCmd.AddCommand(crinitinit.NewCmdInit(out, clientcmd.NewDefaultPathOptions(), defaultServerImage, defaultEtcdImage))
 
-func runHelp(cmd *cobra.Command, args []string) {
-	cmd.Help()
+	return rootCmd
 }
--- a/pkg/crinit/util/util.go
+++ b/pkg/crinit/util/util.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// DefaultClusterRegistryNamespace is the default namespace in which
-	// cluster registry cmponents are hosted.
+	// cluster registry components are hosted.
 	DefaultClusterRegistryNamespace = "clusterregistry"
 )
 


pkg/crinit/crinit.go
```diff
@@ -28,21 +29,15 @@ import (
...
-	cmds := &cobra.Command{
+	rootCmd := &cobra.Command{
 		Use:   "crinit",
 		Short: "crinit runs a cluster registry in a Kubernetes cluster",
 		Long:  "crinit bootstraps and runs a cluster registry as a Deployment in an existing Kubernetes cluster.",
-		Run:   runHelp,
 	}
 
-	// From this point and forward we get warnings on flags that contain "_" separators
-	cmds.SetGlobalNormalizationFunc(flag.WarnWordSepNormalizeFunc)
```
font: This line is the reason we issue a warning about logging before parsing. It's because we have set the warn function as the normalization function when we attempt to convert flags to use dashes instead of underscores before we've parsed our flags.

pkg/crinit/crinit.go
```diff
@@ -28,21 +29,15 @@ import (
...
+	rootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
```
font: I had actually just made a similar change myself to address the lack of V glogs.

pkg/crinit/crinit.go
```diff
@@ -28,21 +29,15 @@ import (
...
+	rootCmd.SetGlobalNormalizationFunc(apiserverflag.WordSepNormalizeFunc)
```
font: Why not use the warning method any longer?

pkg/crinit/crinit.go
perotinus: Yes, I found this by looking at how other k/ command line tools handle this. It's not the prettiest thing, but it seems to be convention at this point.

pkg/crinit/crinit.go
perotinus: Ah, that makes sense. In any case, it seems like the tweak in `cmd/crinit/crinit.go` is the canonical workaround.

pkg/crinit/crinit.go
perotinus: It looks like there are some flags that `glog` adds that have underscores, and this causes a warning to be printed out that there's not much we can do about.

That said, the other k8s tools don't run into this, so there's probably a better way around it.

perotinus:
@font I'd rather not vendor in k/k unless it provides significant additional value. I don't think the `kubectl` Factory API is better enough to be worth the overhead of depending on k/k, but if there are lots of other benefits it brings it may be worthwhile in the longer term.

pkg/crinit/crinit.go
font: The way `kubefed` handles it - or rather how it's handled in the `kubectl` Factory API that `kubefed` relies on - is that it sets the non-warn `WordSepNormalizeFunc` function first, then adds the Go flag set to pflags, then sets the warn `WarnWordSepNormalizeFunc` function. I've seen others just set the non-warn function. I think I prefer to see the warnings and just suppress them for the glog flags.

font:
@perotinus Good point - definitely do not want to vendor in k/k! I guess now I question why `kubectl` isn't split out into its own repo, or at least the useful/common kube command APIs. Perhaps that's already in the works.

cmd/crinit/crinit.go
```diff
@@ -31,6 +32,9 @@ const (
 )
 
 func main() {
+	// Workaround for "ERROR: logging before flag.Parse".
+	flag.CommandLine.Parse([]string{})
```
font: We can remove this based on my comment below.

perotinus: Can you explain this to me in more detail? I've made changes to the normalization function setup, but I still don't see how I can remove this line and avoid `ERROR: logging before flag.Parse`.

font: Oh, sure! My comment is that I think we still want warnings, just not for the ones we already know about (glog). That is, removing this line will just avoid the warnings from the glog flags having underscores. But I think we should still be warned if we've inadvertently added a different flag with an underscore later on in the code. This is achieved because we have set the warning normalization function **after** we've normalized the known glog flags that contain underscores.

perotinus: Ah, I see. In that case, I don't think we can remove this line, since it works around a different issue. I've done the fixes necessary to get the warnings for the normalization function not to apply to the known flags, but I noticed that when running `crinit init` without this line, I still get the `ERROR: logging before flag.Parse` line. There might be another way to remove this, but I couldn't figure it out from examining the code for `kubectl`.

font: Are you 100% sure you're still seeing that error? I applied this patch and removed that line and I don't see any errors.

font: It looks like the reason `kubectl` doesn't run into this problem is because it's using a much older version of glog (based off the glog commit [d1c4472bf2efd3826f2b5bdcc02d8416798d678c](https://github.com/golang/glog/commit/d1c4472bf2efd3826f2b5bdcc02d8416798d678c)) that did not have [this error printed to stderr](https://github.com/golang/glog/blob/23def4e6c14b4da8ac2ed8007337bc5eb5007998/glog.go#L680).

I've seen other workarounds pull in glog based off the commit [44145f04b68cf362d9c4df2182967c2275eaefed](https://github.com/golang/glog/commit/44145f04b68cf362d9c4df2182967c2275eaefed) which is the commit right before that error was introduced by commit [65d674618f712aa808a7d0104131b9206fc3d5ad](https://github.com/golang/glog/commit/65d674618f712aa808a7d0104131b9206fc3d5ad).

So we could either do one of the above. Or if we want to keep a workaround, we can probably just call the golang `flag.Parse()` after all flags are defined and before flags are accessed by the program (per the docs) in `pkg/crinit/crinit.go` right before we return `rootCmd`.

perotinus: Unfortunately, calling flag.Parse() just before returning `rootCmd` short-circuits `cobra`'s parsing of the `-h` flag and means that the standard go flag library prints out just the flags rather than the `cobra` command help.

Sigh. I guess pinning to the same glog version as kubernetes is reasonable. I'd rather not do that in this PR with all the other work going on, so I'm going to remove the fix for that and file an issue to address it after all the vendoring stuff goes in.

perotinus: Filed #66

perotinus:
Thanks! I think not having the workaround is a good reminder for now that this issue needs to be fixed, and it's not a huge usability problem.

PR #61, opened 2017-11-01
 Make necessary changes to add generated code to the repo.

perotinus:
/assign @ericchiang @pmorie @madhusudancs 
/cc @font


docs/development.md
```diff
@@ -91,3 +89,12 @@ some warts.
 1.  When sending out a PR, please put the handmade changes in one commit and the
     mechanical updates in another commit so that it's easier for reviewers to
     see what's been done.
+
+## Updating generated code
+
+If you modify any files in `pkg/apis`, you will likely need to regenerate the
+generated clients and other generated files.
+
+1.  Run ./hack/update-codegen.sh to update the files.
```
madhusudancs: Quote `./hack/update-...` with backticks.

hack/update-codegen.sh
```diff
@@ -74,36 +80,52 @@ function generate_group() {
...
+    --output-base "${OUTPUT_BASE}" \
+    --output-file-base zz_generated.conversion \
+    "$4"
 }
 
+# Set up the temporary GOPATH with necessary dependencies.
 mkdir -p "${TMP_GOPATH}/src/k8s.io/cluster-registry"
 mkdir -p "${TMP_GOPATH}/src/k8s.io/apimachinery"
 cp -r "${SCRIPT_ROOT}/../"* "${TMP_GOPATH}/src/k8s.io/cluster-registry"
 cp -r "${SCRIPT_ROOT}/../vendor/k8s.io/apimachinery/"* "${TMP_GOPATH}/src/k8s.io/apimachinery"
```
font: I know this is from a past commit, but what's the reason for copying over apimachinery into the temporary GOPATH?

perotinus: `client-gen` requires that the apimachinery repo be in the GOPATH. I'm not sure exactly why that dependency exists, but client generation fails without some packages in `apimachinery`.

perotinus:
Thanks for the reviews!

hack/update-codegen.sh
font: Ah ok, thanks for the explanation. Maybe just add a comment about that next time you're in this area.

PR #62, opened 2017-11-01
 Modify the cluster-registry repo to enable checked-in BUILD files

perotinus:
This should complete the work necessary for #24 and unblock vendoring of this repo.

Only the last two commits (starting with "Modify the cluster-registry repo to enable checked-in BUILD files." are meant to be reviewed here). See #61 for the reviews of the other commits.

/assign @madhusudancs @pmorie @ericchiang 
/cc @font


hack/verify-bazel.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
font: Is this just a handy script to have or are there other plans for this script e.g. CI? We may want to consider adding it to the development doc.

perotinus: Yes, the idea is to be able to run it in CI.

Agreed that it should be added to the docs. I think I'd like to do a doc fixup after all these PRs are in, rather than updating the docs multiple times in the middle of reviews, if that's OK with you.

Issue #63, opened 2017-11-02
Modify crinit to deploy cluster registry using k8s apiserver aggregation layer


font:
This tracks the work to modify `crinit` to deploy the cluster registry using API aggregation. See https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/ for required steps.

font:
I've started working on this.

PR #65, opened 2017-11-02
Add support to crinit to deploy clusterregistry using k8s API aggregator

font:
In support of #63.


pkg/crinit/init/init.go
```diff
@@ -120,11 +122,12 @@ func (o *initClusterRegistryOptions) Bind(flags *pflag.FlagSet, defaultServerIma
...
+	flags.StringVar(&o.apiServerServiceTypeString, apiserverServiceTypeFlag, string(v1.ServiceTypeNodePort), "The type of service to create for the cluster registry. Options: 'LoadBalancer', 'NodePort'.")
 	flags.StringVar(&o.apiServerAdvertiseAddress, apiserverAdvertiseAddressFlag, "", "Preferred address at which to advertise the cluster registry API server NodePort service. Valid only if '"+apiserverServiceTypeFlag+"=NodePort'.")
 	flags.Int32Var(&o.apiServerNodePortPort, apiserverPortFlag, 0, "Preferred port to use for the cluster registry API server NodePort service. Set to 0 to randomly assign a port. Valid only if '"+apiserverServiceTypeFlag+"=NodePort'.")
 	flags.BoolVar(&o.apiServerEnableHTTPBasicAuth, "apiserver-enable-basic-auth", false, "Enables HTTP Basic authentication for the cluster registry API server. Defaults to false.")
 	flags.BoolVar(&o.apiServerEnableTokenAuth, "apiserver-enable-token-auth", false, "Enables token authentication for the cluster registry API server. Defaults to false.")
+	flags.BoolVar(&o.apiServerStandalone, apiServerStandaloneFlag, false, "Disables the use of the Kubernetes API aggregation layer")
```
perotinus: I'm not sure what the right way to think about this flag is. I think there are potentially multiple things that are involved with standalone vs aggregated:

- service type
- authentication model
- necessary flags
- necessary backend setup
- more that I haven't thought of

I wonder if it would make sense to make two separate commands `initaggreated` and `initstandalone` (with better names) that could share logic but have separate flags as necessary.

pkg/crinit/init/init.go
```diff
@@ -411,17 +417,20 @@ func createService(cmdOut io.Writer, clientset client.Interface, namespace, svcN
 
 	ips := []string{}
 	hostnames := []string{}
-	if apiserverServiceType == v1.ServiceTypeLoadBalancer {
-		ips, hostnames, err = waitForLoadBalancerAddress(cmdOut, clientset, svc, dryRun)
-	} else {
-		if apiserverAdvertiseAddress != "" {
-			ips = append(ips, apiserverAdvertiseAddress)
+
```
perotinus: I think this logic might be simpler to read with a short return:

```
if !apiServerStandalone {
  return svc, []string{}, []string{}, err
}

ips := []string{}
hostnames := []string{}

if apiserverServiceType == ...
...
```

pkg/crinit/init/init.go
font: Yes, after I made this flag I started realizing there's going to be a bit of flag error checking and confusing list of possible arguments. I think limiting that with new subcommands would help.

The clusterregistry is going to have to deal with similar problems. We probably also want to think about how the clusterregistry should handle the differing arguments. Introducing new subcommands for an apiserver seems strange. So does maintaining 2 separate binaries.

font: I opted for `crinit aggregated` and `crinit standalone` since the executable is already named `crinit`. Do we envision any other uses for this binary other than to initialize a CR? If that time comes we could then modify the command names too.

perotinus:
I'll take a closer look at this later, but I think this looks like a good direction.

Some high-level comments:
- I think we can share more code between the commands, but I'm OK deferring that if you're willing to commit to a cleanup round later once these are both fleshed out. It could be the case that they diverge enough that less code is shareable than would now appear to be the case.
- I think some tests need to migrate
- I think the "standalone" and "aggregated" naming works for me for now. I'm assuming that there is some explanation of what they are if you run just `crinit`?

font:
> - I think we can share more code between the commands, but I'm OK deferring that if you're willing to commit to a cleanup round later once these are both fleshed out. It could be the case that they diverge enough that less code is shareable than would now appear to be the case.

Yes, we can definitely share more code and that's the direction I plan to go in. This is just WIP to provide a peek into the direction I'm heading so that I don't waste too much effort exploring ideas.

> - I think some tests need to migrate

Agreed as this is WIP, we should have `aggregated_test.go` similar to `standalone_test.go`.

> - I think the "standalone" and "aggregated" naming works for me for now. I'm assuming that there is some explanation of what they are if you run just crinit?

Yes, that's the idea. But I will plan to add more details on what the naming means specifically in their respective long explanations.

After going through some of this refactoring exercise, it doesn't really seem like there are too many unique flags to each of the subcommands. The only uniqueness right now seems to be the authentication methods for standalone vs aggregated not having any. For example, while at first service type seemed to be one of those unique flags, hypothetically speaking, you could have a load balancer service type even in aggregated mode. Although that may not be very common. But this makes both deployment options support both LoadBalancer and NodePort - perhaps with a different default for each.

In any case, I still think it's clearer to split out the subcommands and the future may have more mutually exclusive options. But I wanted to mention it nonetheless.

perotinus:
I agree that it's clearer to split out the subcommands, even if they are currently very similar. Code wise, I believe that they are likely to diverge, and having two separate files to demonstrate that divergence rather than a series of branches in one file.

One thing we may want to consider: there may be a point when we want to add a `delete` or `modify` command at the root, in which case it may be a bit odd to have `standalone` and `aggregate` at the top level of the command. That said, calling it `crinit` will also be an issue at that point.

font:
Yep, I agree. We could then consider renaming it to `kubecr` or similar and renaming the sub-commands appropriately.

perotinus:
Filed an issue to track separately.

perotinus:
@font is this ready for review?

font:
@perotinus Not yet - I'll update the PR title to remove WIP and post a comment about it when it's ready.

pkg/crinit/standalone/standalone_test.go
```diff
@@ -28,63 +28,94 @@ import (
...
+			initialOpts: &standaloneClusterRegistryOptions{
+				apiServerServiceTypeString: string(v1.ServiceTypeNodePort),
+				SubcommandOptions: util.SubcommandOptions{
+					ApiServerAdvertiseAddress: "10.0.0.1"}},
 			errExpected: false,
 		},
 		{
-			desc:        "adveritse address not supported with non-NodePort service type",
-			initialOpts: &initClusterRegistryOptions{apiServerServiceTypeString: string(v1.ServiceTypeLoadBalancer), apiServerAdvertiseAddress: "10.0.0.1"},
+			desc: "adveritse address not supported with non-NodePort service type",
```
pmorie: typo: advertise

pkg/crinit/standalone/standalone_test.go
```diff
@@ -28,63 +28,94 @@ import (
...
+			initialOpts: &standaloneClusterRegistryOptions{
+				apiServerServiceTypeString: string(v1.ServiceTypeLoadBalancer),
+				SubcommandOptions: util.SubcommandOptions{
+					ApiServerAdvertiseAddress: "10.0.0.1"}},
 			errExpected: true,
 		},
 		{
-			desc:        "adveritse address validated",
-			initialOpts: &initClusterRegistryOptions{apiServerServiceTypeString: string(v1.ServiceTypeNodePort), apiServerAdvertiseAddress: "notAValidIP"},
+			desc: "adveritse address validated",
```
pmorie: typo: advertise

gnirodi:
Hello, github.com/istio/istio has also been exploring using the cluster registry as an aggregated API. Taking a brief look at the PR it's possible we could adopt it as is. 

1. Could you point us to any doc that describes how to set this up?
2. Is the expectation that this is always a standalone binary or is there the possibility of another binary using this capability as a library. For the latter, the cluster registry would still be run as a  distinct endpoint.
3. How far along is this PR in terms of becoming part of the repro?

font:
>1. Could you point us to any doc that describes how to set this up?

The only documentation we have so far is our [development doc](https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md). Unfortunately, there are no user docs for this yet. We are working towards an alpha release of the cluster registry and are tracking user documentation in #89. I can provide some pointer commands that I've been using in my development of this PR if you'd like. The commands are slightly different from what's currently in `master@HEAD` because this PR is changing the subcommands to support standalone and aggregated deployments.

>2. Is the expectation that this is always a standalone binary or is there the possibility of another binary using this capability as a library. For the latter, the cluster registry would still be run as a distinct endpoint.

The standalone binary is the initial use case we're solving. If projects want to vendor the cluster registry then they will be able to do so; we plan to support that and this repo should now support being vendored. To support the cluster registry being consumed as a library, however, has not been discussed AFAIK.

>3. How far along is this PR in terms of becoming part of the repro?

It's getting close. The `crinit` changes to deploy the cluster registry using the aggregator is mostly complete. But the `clusterregistry` apiserver changes still need to be done to support delegated authn/authz.

perotinus:
@font I see there's been some progress. Is this still WIP, or should I review?

font:
@perotinus Still WIP. Besides a little cleanup, I think most of the `crinit` changes are done though. Now the `apiserver` will need some modifications, I may do a separate PR for that as this PR is starting to get a little unwieldy. But first, I need consensus on best approach for #69.

font:
@perotinus @pmorie This is now ready to review. This only includes `crinit` changes. I will open up another PR for the `clusterregistry` apiserver changes.

Gopkg.toml
```diff
@@ -77,3 +77,7 @@ ignored = ["github.com/kubernetes/repo-infra/kazel"]
 [[override]]
   name = "k8s.io/gengo"
   revision = "75356185a9af8f0464efa792e2e9508d5b4be83c"
+
+[[constraint]]
+  branch = "master"
```
perotinus: Does it make sense to pin this to the same version that k/k uses?

pkg/crinit/standalone/standalone.go
```diff
@@ -0,0 +1,185 @@
...
+func validateOptions(opts *standaloneClusterRegistryOptions) error {
+	opts.ApiServerServiceType = v1.ServiceType(opts.apiServerServiceTypeString)
+	return opts.ValidateCommonOptions()
+}
+
+// Run initializes a cluster registry.
+func Run(opts *standaloneClusterRegistryOptions, cmdOut io.Writer,
+	hostClientset client.Interface, pathOptions *clientcmd.PathOptions) error {
+
+	// Create Cluster Registry Namespace
```
perotinus: These inline comments might be TMI: the method names you've chosen are pretty clear.

pkg/crinit/standalone/standalone.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	flags := cmd.Flags()
+	opts.BindCommon(flags, defaultServerImage, defaultEtcdImage)
+	opts.Bind(flags)
+
+	return cmd
+}
+
+type entityKeyPairs struct {
```
perotinus: Are these still needed here?

pkg/crinit/standalone/standalone_test.go
```diff
@@ -177,7 +210,7 @@ func TestCreateNamespace(t *testing.T) {
 	t.Run("simple namespace creation", func(t *testing.T) {
 		name := "test"
 		client := fake.NewSimpleClientset()
-		ns, err := createNamespace(client, name, false)
+		ns, err := util.CreateNamespace(client, name, false)
```
perotinus: This seems like it should be moved into a util_test rather than here in standalone_test.

pkg/crinit/util/util.go
```diff
@@ -18,33 +18,138 @@ package util
...
 )
 
 // SubcommandOptions holds the configuration required by the subcommands of
 // `clusterregistry`.
 type SubcommandOptions struct {
-	Name                     string
-	Host                     string
-	ClusterRegistryNamespace string
-	Kubeconfig               string
+	Name                      string
```
perotinus: I wonder if it makes sense to create a specific options package. This file has become very large to be called `util`.

I could see three packages:
- `options`, for options and related
- `common` (ideally with a better name), for functionality that's shared between `standalone` and `aggregated`
- `util`, for things that are specifically utilities

pkg/crinit/util/util.go
```diff
@@ -53,6 +158,788 @@ func (o *SubcommandOptions) SetName(args []string) error {
...
+	glog.V(4).Infof("Created service named %s with IP addresses %v, hostnames %v",
+		svc.Name, ips, hostnames)
+
+	return svc, ips, hostnames, err
+}
+
+// CreateService helper to create the cluster registry apiserver service object
+// and return the object. If service type is load balancer, will wait for load
+// balancer IP and return it and hostnames
+func CreateService(cmdOut io.Writer, clientset client.Interface, namespace,
```
perotinus: Should this be private?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,412 @@
...
+	aggregated_example = `
+	# Initialize a cluster registry named foo
+	# in the host cluster whose local kubeconfig
+	# context is bar.
+	crinit aggregated foo --host-cluster-context=bar`
+
+	apiServiceGroupPriorityMinimum int32 = 10000
+	apiServiceVersionPriority      int32 = 20
+
+	clusterRegistryAPIGroup         = v1alpha1.GroupName + ".k8s.io"
```
perotinus: I think this should change for the clusterregistry API rather than just here: it's group name should be `clusterregistry.k8s.io`

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,412 @@
...
+	clusterRoleBindingAuthDelegator = v1alpha1.GroupName + ".k8s.io:apiserver-auth-delegator"
+
+	clusterRoleAPIGroup  = []string{v1alpha1.GroupName}
+	clusterRoleResources = []string{"cluster"}
+	clusterRoleVerbs     = []string{"get", "list", "watch", "create", "update", "patch", "delete"}
+)
+
+type aggregatedClusterRegistryOptions struct {
+	util.SubcommandOptions
+	apiServerServiceTypeString string
```
perotinus: Isn't this also an option to the standalone server?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	// Create RBAC objects
+	err = createRBACObjects(cmdOut, hostClientset, opts)
+	if err != nil {
+		return err
+	}
+
+	// Update kubeconfig
+	err = opts.UpdateKubeconfig(cmdOut, pathOptions, svc, ips, hostnames,
```
perotinus: Why do you need to do this? Isn't the idea to access the cluster registry via the kube-apiserver?

Gopkg.toml
font: It would, except k/k does not vendor `kube-aggregator`. It's just used as an import. k/f also uses this strategy i.e. HEAD at the time of vendoring.

pkg/crinit/aggregated/aggregated.go
font: Yes, except that I've opted to make `NodePort` the default for aggregated and left `LoadBalancer` the default for standalone.

pkg/crinit/aggregated/aggregated.go
font: Agreed - fixed!

pkg/crinit/aggregated/aggregated.go
font: Yes, although I chose to leave this in because hypothetically you may want to talk directly to the `clusterregistry` API server. For example, I still left the `LoadBalancer` API server service type for this same reason. Thoughts?

pkg/crinit/standalone/standalone.go
font: Nope, removed.

pkg/crinit/standalone/standalone.go
font: Removed.

pkg/crinit/standalone/standalone_test.go
font: Yes, when I follow up with an aggregated_test I'll consolidate for sure.

pkg/crinit/util/util.go
font: Agreed, but for the sake of increasing the size of this PR I left as is. I can follow up with another PR.

pkg/crinit/util/util.go
font: It is used in `standalone_test.go`.

Gopkg.toml
perotinus: Got it, thanks for the explanation.

pkg/crinit/aggregated/aggregated.go
perotinus: Oh, I see. In that case, I think this is fine.

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+
+var (
+	aggregated_long = `
+	Aggregated initializes an aggregated cluster registry.
+
+	The aggregated cluster registry is hosted inside a Kubernetes
+	cluster and registers its API with the Kubernetes API aggregator.
+	The host cluster must be specified using the --host-cluster-context flag.`
+	aggregated_example = `
+	# Initialize a cluster registry named foo
```
perotinus: Perhaps update this documentation to specifically talk about aggregation?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	# in the host cluster whose local kubeconfig
+	# context is bar.
+	crinit aggregated foo --host-cluster-context=bar`
+
+	apiServiceGroupPriorityMinimum int32 = 10000
+	apiServiceVersionPriority      int32 = 20
+
+	clusterRegistryAPIService       = v1alpha1.SchemeGroupVersion.Version + "." + v1alpha1.GroupName
+	serviceAccount                  = v1alpha1.GroupName + "-apiserver"
+	clusterRole                     = v1alpha1.GroupName + ".k8s.io:apiserver"
```
perotinus: I think these need to be updated now that the group name is `clusterregistry.k8s.io`

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	aggregated_example = `
+	# Initialize a cluster registry named foo
+	# in the host cluster whose local kubeconfig
+	# context is bar.
+	crinit aggregated foo --host-cluster-context=bar`
+
+	apiServiceGroupPriorityMinimum int32 = 10000
+	apiServiceVersionPriority      int32 = 20
+
+	clusterRegistryAPIService       = v1alpha1.SchemeGroupVersion.Version + "." + v1alpha1.GroupName
```
perotinus: Perhaps you could add comments here as to what all these roles are and what they allow? It would at least help later when someone is writing user documentation.

pkg/crinit/aggregated/aggregated.go
perotinus: Hmm. IMO it would be better to make this simpler and let people who want to access the cluster registry directly from kubectl do that themselves. I can't think of a use case that necessitates this, and I'd rather make this code simpler to start and complicate it when there's a user need.

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = opts.CreateAPIServerCredentialsSecret(hostClientset, credentials)
+	if err != nil {
+		return err
+	}
+
+	var pvc *v1.PersistentVolumeClaim
```
perotinus: Why declare this? Can't you use `:=` like above?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return err
+	}
+
+	return opts.WaitForAPIServer(cmdOut, hostClientset, pathOptions, ips,
+		hostnames, svc)
+}
+
+// createRbacObjects handles the creation of all the RBAC objects necessary
+// to deploy the cluster registry in aggregated mode
```
perotinus: nit: end with a period.

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+		credentials)
+	if err != nil {
+		return err
+	}
+
+	return opts.WaitForAPIServer(cmdOut, hostClientset, pathOptions, ips,
+		hostnames, svc)
+}
+
+// createRbacObjects handles the creation of all the RBAC objects necessary
```
perotinus: nit: createRBACObjects

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+
+	if err != nil {
+		glog.V(4).Infof("Failed to create service account %v: %v", sa, err)
+		return err
+	}
+
+	glog.V(4).Info("Successfully created service account")
+
+	// Create a Kubernetes cluster role for the operations you want to allow
+	// on your resources
```
perotinus: nit: This comment seems pretty generic. Can you make it more specific to what you're doing here for the cluster registry?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+
+	fmt.Fprintln(cmdOut, " done")
+	glog.V(4).Info("Successfully created cluster role binding")
+	return nil
+}
+
+// createServiceAccount handles the creation of the service account for the
+// the cluster registry to be used with RBAC.
+func createServiceAccount(cmdOut io.Writer, clientset client.Interface,
+	opts *aggregatedClusterRegistryOptions) (*v1.ServiceAccount, error) {
```
perotinus: I think it would be better to just pass `name` and `dryRun`: it makes the coupling here looser. Consider that for other method calls too, though I can understand if you decide that the parameter lists are too long

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	fmt.Fprintln(cmdOut, " done")
+	glog.V(4).Info("Successfully created cluster role binding")
+	return nil
+}
+
+// createServiceAccount handles the creation of the service account for the
+// the cluster registry to be used with RBAC.
+func createServiceAccount(cmdOut io.Writer, clientset client.Interface,
```
perotinus: Consider not massing cmdOut unless you are using it.

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	if opts.DryRun {
+		return cr, nil
+	}
+
+	return clientset.RbacV1().ClusterRoles().Create(cr)
+}
+
+// createClusterRoleBinding creates the cluster role binding for the operations we will allow
```
perotinus: nit: createClusterRoleBindings

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	// cluster role binding for clusterregistry.k8s.io:apiserver cluster role
+	crb1 := &rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   clusterRoleBindingAPIServer,
+			Labels: util.ComponentLabel,
+		},
+		Subjects: []rbacv1.Subject{
+			rbacv1.Subject{
+				Kind:      "ServiceAccount",
+				APIGroup:  "",
```
perotinus: There's no API group? Can you just omit this field?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     clusterRole,
+		},
+	}
+
+	// cluster role binding for system:auth-delegator cluster role
+	crb2 := &rbacv1.ClusterRoleBinding{
```
perotinus: Maybe authDelegatorCRB (or authDelegatorRoleBinding)?

And call the previous one apiServerCRB?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+		},
+	}
+
+	if opts.DryRun {
+		return crb1, crb2, nil, nil
+	}
+
+	result1, err1 := clientset.RbacV1().ClusterRoleBindings().Create(crb1)
+	result2, err2 := clientset.RbacV1().ClusterRoleBindings().Create(crb2)
+	return result1, result2, err1, err2
```
perotinus: I think it's a more standard pattern to return if the first call fails. So, this method would have a return signature `(*rbacv1.ClusterRoleBinding, *rbacv1.ClusterRoleBinding, error)`

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		glog.V(4).Infof("Failed to create Kubernetes API Service %v: %v", apiSvc, err)
+		return nil, err
+	}
+
+	glog.V(4).Info("Successfully created Kubernetes API Service")
+	return apiSvc, nil
+}
+
+// createAPIServiceObject helper to create the cluster registry API Service
```
perotinus: nit: createAPIServiceObject creates and returns the cluster registry API Server object.

I don't think you need to specify that this is a helper.

pkg/crinit/util/util.go
```diff
@@ -18,33 +18,138 @@ package util
...
+	// to bind to port < 1000.  The apiserver's service will still
+	// expose on port 443.
+	apiServerSecurePort = 8443
+)
+
+var (
+	serverName     string
+	serverCredName string
+
+	ComponentLabel = map[string]string{
```
perotinus: Why is this one public?

pkg/crinit/util/util.go
perotinus: SGTM.

pkg/crinit/util/util.go
```diff
@@ -53,6 +158,788 @@ func (o *SubcommandOptions) SetName(args []string) error {
...
+		argsMap[key] = val
+	}
+
+	o.ApiServerOverrides = argsMap
+
+	return nil
+}
+
+// CreateNamespace creates the cluster registry namespace
+func (o *SubcommandOptions) CreateNamespace(cmdOut io.Writer,
```
perotinus: I'm not sure how I feel about having these methods be on the SubcommandOptions object: it seems a bit odd to call into an options object to do logic like this.

pkg/crinit/util/util.go
perotinus: Ah, I see. I assume that will change in the follow-up, when the standalone tests are moved to util.

pmorie:
@font rebase needed on this one

pkg/crinit/aggregated/aggregated.go
perotinus: Please fix before merging.

perotinus:
@font @pmorie There's one comment above that I think is a functional issue, but everything else can wait for a follow-up. If you rebase, I'll approve and we can merge.

font:
@perotinus @pmorie Updated based on comments.

pkg/crinit/aggregated/aggregated.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = createRBACObjects(cmdOut, hostClientset, opts)
+	if err != nil {
+		return err
+	}
+
+	err = opts.UpdateKubeconfig(cmdOut, pathOptions, svc, ips, hostnames,
```
perotinus: Couldn't find the old comment thread here, so starting a new one.

I think you should remove this for now. It's easier to add it back later if people want this functionality, and I'm not sure why a user would want to interact with the cluster registry like this if they're aggregating it.

Sorry, I thought I had made that comment already in the last pass!

perotinus:
Thanks for the fixes! Just one more comment, then I think this is ready to merge.

pkg/crinit/aggregated/aggregated.go
font: Yeah I saw this comment but I couldn't reply to it.

I actually did try to remove it but ran into an error caused by the subsequent call to `WaitForAPIServer`. Since that method is shared by both standalone and aggregated I opted to worry about it later since it doesn't affect the functionality. So I don't disagree, but would prefer to follow up with another PR. WDYT?

perotinus: SGTM.

perotinus:
@font Can you squash this into two commits, one for the changes to `Gopkg.toml` and the `vendor/` directory, and one for everything else? I'll merge it once that's done.

perotinus:
Thanks @font!

PR #72, opened 2017-11-09
Update the glog dependency version to match the version used in Kubernetes

### LINKED ISSUE #66: Fix ERROR: logging before flag.Parse in crinit
Based on comments in #60, we plan to revert to the glog version used by k8s. However, because of all the outstanding repo infra work (#58, #61, #62) I want to defer this so as not to complicate those changes.

perotinus:
Fixes #66.

/cc @madhusudancs 
fyi @font 

/sig multicluster

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -146,7 +146,7 @@
 [[projects]]
   name = "github.com/golang/glog"
   packages = ["."]
-  revision = "23def4e6c14b4da8ac2ed8007337bc5eb5007998"
+  revision = "44145f04b68cf362d9c4df2182967c2275eaefed"
 
 [[projects]]
   branch = "master"
@@ -448,6 +448,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "c1b65cf753e556bc4f8dfee24ca2beccc73118b8f62bed27ea54f2e9430d601c"
+  inputs-digest = "0e95da9a54b18849cff5008263c4b7654162e367be977c46219af7b1a0b14055"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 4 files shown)


Gopkg.lock
```diff
@@ -146,7 +146,7 @@
 [[projects]]
   name = "github.com/golang/glog"
   packages = ["."]
-  revision = "23def4e6c14b4da8ac2ed8007337bc5eb5007998"
+  revision = "44145f04b68cf362d9c4df2182967c2275eaefed"
```
font: I see that this sha revision comes from https://github.com/kubernetes/kubernetes/blob/f34d3af809169007e7f474a26751ee675c0664ac/Godeps/Godeps.json#L1335. Do you know how it relates to the version in the vendor directory https://github.com/kubernetes/kubernetes/blame/master/vendor/github.com/golang/glog/glog.go?

madhusudancs: @font that threw me off too, but at this point I wouldn't read too much into our (k/k's) vendor directory's history or do any sort of software archaeology using git tools. They are all more or less useless at high level today, unless you are willing to spend hours trying to figure out how things have moved around. This is because of the way the dependencies have moved from fully-Godeps managed to who-knows-what-we-use-today (hand-managed at best?) for dependency management.

The version you are pointing to in the k/k repo with the blame UI is at the same version Jonathan is updating this dependency to. Either look at https://github.com/kubernetes/kubernetes/commits/899f9b4e31218172eafcf0a5de447e730619fbbe/Godeps/_workspace/src/github.com/golang/glog/glog.go or scroll down the blame page observing all the updates that are next to other lines in the file. The latest update is recorded at https://github.com/kubernetes/kubernetes/blame/master/vendor/github.com/golang/glog/glog.go#L318 which updates glog in `k/k/Godeps/_workspace` to `44145f0` and then at some point `k/k/Godeps/_workspace` is moved to `k/k/vendor`.

font: @madhusudancs Thanks for the explanation! It definitely seems like a bit of a rabbit hole.

PR #73, opened 2017-11-09
Remove the hack/bazel-test.sh script.

perotinus:
/sig multicluster

Submit after https://github.com/kubernetes/test-infra/pull/5433.


perotinus:
Submitting. I don't think this is a contentious change, and the passing test is all the necessary verification.

PR #76, opened 2017-11-16
Ignore .vscode folder in project root

pmorie:
E-Z review

/sig multicluster

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /bazel-*
+.vscode


.gitignore
```diff
@@ -1 +1,2 @@
 /bazel-*
+.vscode
```
font: How about a short comment describing this?

perotinus:
I'm inclined to agree with @font here. This seems like it's more appropriate for `.git/info/exclude`.

pmorie:
SGTM

Issue #102, opened 2017-11-17
Get a nightly build of the image and the tool going


perotinus:
/sig multicluster

It would be good to have a nightly image to make it easier for people to test. This could perhaps be done with a prow job for now, and perhaps later migrate to using Cloud Container Builder.

perotinus:
I'd like to have this by kubecon if possible, so that people have an easy experimentation path.

perotinus:
I've followed #111 to track cross compilation.

perotinus:
Prow changes are in. Will close this tomorrow if it succeeds this evening.

perotinus:
https://k8s-testgrid.appspot.com/sig-multicluster-cluster-registry#nightly succeeded last night.

PR #109, opened 2017-11-30
Add a script that does a nightly build of the cluster registry.

perotinus:
/sig multicluster

Addresses #102, though there's still some more clean-up work necessary to make the the user experience smooth.

/cc @font @pmorie @madhusudancs

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -57,7 +57,9 @@ To build it, from the root of the repository:
 1.  If you want to build a docker image, run
     `bazel build //cmd/clusterregistry:clusterregistry-image`
 1.  To push an image to Google Container registry, you'll need to run
-    `bazel run //cmd/clusterregistry:push-clusterregistry-image --define project=<your_project_id>`
+    `bazel run //cmd/clusterregistry:push-clusterregistry-image --define repository=<your_gcr_repository_path>`
+     where `your_gcr_repository_path` is your GCP project name followed by
+     your image name, e.g., `myproject/myimage`.
 
 ## Building `crinit`
 
@@ -111,5 +113,51 @@ If you modify any files in `pkg/apis`, you will likely need to regenerate the
 generated clients and other generated files.
 
 1.  Run `./hack/update-codegen.sh` to update the files.
-1.  Add the generated files to your PR, preferably in a separate,
-    generated-only commit so that they are easier to review.
+1.  Add the generated files to your PR, preferably in a separate, generated-only
+    commit so that they are easier to review.
+
+## Nightly releases
+
+The cluster registry has a script, [`hack/release.sh`](../hack/release.sh), that
+is used to build releases and push them for public consumption. This script is
+run nightly by Prow.
+
+### Binaries
+
+Binaries are stored in Google Cloud Storage, in the `crreleases` bucket.
+Currently there are only binaries releases for 64-bit Linux.
+
+To get the latest client library (i.e., `crinit`), run:
+
+```sh
+PACKAGE=client
+LATEST=$(curl https://storage.googleapis.com/crreleases/nightly/latest)
+curl -O http://storage.googleapis.com/crreleases/nightly/$LATEST/clusterregistry-$PACKAGE.tar.gz
+```
+
+To verify, run:
+
+```sh
+curl http://storage.googleapis.com/crreleases/nightly/$LATEST/clusterregistry-$PACKAGE.tar.gz.sha | sha256sum -c -
+```
+
+To get the latest server binaries, run the commands above but replace
+`PACKAGE=client` with `PACKAGE=server`.
+
+### Images
+
+A Docker image for the [`clusterregistry`](../cmd/clusterregistry) binary is
+pushed to GCR.
+
+To pull the latest image, run:
+
+```sh
+docker pull gcr.io/crreleases/nightly/clusterregistry:latest_nightly
+```
+
+To pull an image from a specific date, replace `latest` with the date in
+`YYYYMMDD` format. For example,
+
+```sh
+docker pull gcr.io/crreleases/nightly/clusterregistry:20171201
+```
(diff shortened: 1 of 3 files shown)


docs/development.md
```diff
@@ -111,5 +113,51 @@ If you modify any files in `pkg/apis`, you will likely need to regenerate the
...
+
+Binaries are stored in Google Cloud Storage, in the `crreleases` bucket.
+Currently there are only binaries releases for 64-bit Linux.
+
+To get the latest client library (i.e., `crinit`), run:
+
+```sh
+PACKAGE=client
+LATEST=$(curl https://storage.googleapis.com/crreleases/nightly/latest)
+curl -O http://storage.googleapis.com/crreleases/nightly/$LATEST/clusterregistry-$PACKAGE.tar.gz
```
madhusudancs: It's not necessary at this pre-alpha stage where everything is expected to change, but in the near future we need to start including platform (linux) and arch (amd64) in the URL path to not break the paths in the URLs.

docs/development.md
```diff
@@ -111,5 +113,51 @@ If you modify any files in `pkg/apis`, you will likely need to regenerate the
...
+
+### Images
+
+A Docker image for the [`clusterregistry`](../cmd/clusterregistry) binary is
+pushed to GCR.
+
+To pull the latest image, run:
+
+```sh
+docker pull gcr.io/crreleases/nightly/clusterregistry:20171130
```
madhusudancs: You mean s/20171130/latest/ here?

I would rather call this tag `nightly` and reserve `latest` for the latest version of a stable release.

hack/release.sh
```diff
@@ -0,0 +1,60 @@
...
+trap clean_up EXIT
+
+cd "${SCRIPT_ROOT}/.."
+
+# Build copy the tools.
+bazel build //cmd/crinit //cmd/clusterregistry
+
+# Create the archives.
+pushd bazel-bin/cmd/crinit
+tar czf "${TMPDIR}/clusterregistry-client.tar.gz" crinit
```
madhusudancs: Would `tar czf "${TMPDIR}/clusterregistry-client.tar.gz" -C bazel-bin/cmd/crinit crinit` work instead of `pushd`/`popd` headache? What bothers me about `pushd`/`popd` are the errors that leave the users in a `bazel-bin/...` directory.

hack/release.sh
```diff
@@ -0,0 +1,60 @@
...
+popd
+pushd bazel-bin/cmd/clusterregistry
+tar czf "${TMPDIR}/clusterregistry-server.tar.gz" clusterregistry
+popd
+
+pushd "${TMPDIR}"
+
+# Create the SHAs.
+sha256sum clusterregistry-client.tar.gz > clusterregistry-client.tar.gz.sha
```
madhusudancs: Same here. Except use absolute paths?

hack/release.sh
```diff
@@ -0,0 +1,60 @@
...
+sha256sum clusterregistry-server.tar.gz > clusterregistry-server.tar.gz.sha
+
+# Upload the files to GCS.
+gsutil cp * gs://crreleases/${RELEASE_DIR}
+
+popd
+
+# Push the server container image.
+bazel run //cmd/clusterregistry:push-clusterregistry-image --define repository=crreleases/nightly/clusterregistry
+gcloud container images add-tag --quiet \
```
madhusudancs: This is cool! TIL. Didn't know this existed.

hack/release.sh
perotinus: Yes it does!

docs/development.md
perotinus: Yes, I did mean latest.

I think `latest_nightly` makes sense as a tag. I've updated this to match.

hack/release.sh
perotinus: In this case, I don't think that's possible. sha256sum generates a SHA with the full path to the file if you provide it. So, if you run `sha256gen /foo/bar`, the output SHA will reference that folder and will fail to verify.

hack/release.sh
perotinus: Yeah, I'm glad this existed. Hopefully with the Prow service account set up properly this will just work too.

docs/development.md
perotinus: Yes, agreed. Bazel 0.8.0 supports cross compilation, and once that's available, I'll migrate this repo to use the newest version of `rules_go` and update this script to cross compile.

perotinus:
Thanks @madhusudancs! Addressed your comments.

perotinus:
@madhusudancs @pmorie @font If there are no objections, I plan to merge this today and get the nightly build going.

hack/release.sh
```diff
@@ -0,0 +1,65 @@
...
+  gcr.io/crreleases/nightly/clusterregistry:dev \
+  gcr.io/crreleases/nightly/clusterregistry:${BUILD_DATE}
+gcloud container images add-tag --quiet \
+  gcr.io/crreleases/nightly/clusterregistry:dev \
+  gcr.io/crreleases/nightly/clusterregistry:latest_nightly
+gcloud container images untag --quiet \
```
madhusudancs: Why tag and untag `dev`? Could you also add the explanation as a comment?

perotinus: Tagging the image as `dev` makes it easier to add the other tags. If the image didn't have a `dev` tag, it would be more cumbersome to figure out which image to tag. I added a comment to this effect.

pmorie:
Test grid build: https://k8s-testgrid.appspot.com/sig-multicluster-cluster-registry

hack/release.sh
madhusudancs: Understood. Thanks for the explanation.

perotinus:
Test build failed because the service account didn't have the appropriate permissions. I'm fixing that now, and will try to manually run the job.

PR #116, opened 2017-12-01
Update the release script to configure gcloud Docker credentials.

perotinus:
/sig multicluster

This should fix the nightly build.

/cc @madhusudancs @font @pmorie

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -29,6 +29,12 @@ trap clean_up EXIT
 BUILD_DATE="$(TZ=Etc/UTC date +%Y%m%d)"
 RELEASE_DIR="${TMPDIR}/${BUILD_DATE}"
 
+# Check for and install necessary dependencies.
+command -v bazel >/dev/null 2>&1 || { echo >&2 "Please install bazel before running this script."; exit 1; }
+command -v gcloud >/dev/null 2>&1 || { echo >&2 "Please install gcloud before running this script."; exit 1; }
+gcloud components install gsutil docker-credential-gcr
+docker-credential-gcr configure-docker
+
 cd "${SCRIPT_ROOT}/.."
 
 # Build the tools.


hack/release.sh
```diff
@@ -53,6 +53,9 @@ popd
 # Upload the files to GCS.
 gsutil cp -r "${TMPDIR}"/* "gs://crreleases/${ROOT_RELEASE_DIR}"
 
+gcloud components install docker-credential-gcr
```
madhusudancs: Is there a way we can do a pre-flight check for gcloud and its components such as gsutil, docker-credential-gcr etc.? What if gsutil is not installed? What if docker-credential-gcr is already installed? I guess the latter is not a big deal because gcloud silently skips the step if the component already exists, but the former one is a problem. This problem is not introduced by this PR, but I somehow overlooked it before. So I am sorry for missing it.

perotinus: I can do a preflight check for gcloud and bazel. I don't think there's a convenient way to do a preflight check for gcloud components. I can filter the list of components and grep for "Installed", but that seems like an unclean solution when running install for an already-installed component is a no-op. (And I believe specifically designed to be so, since there's a `reinstall` command as well.)

madhusudancs: How about just running a pre-flight install then? For example, just run `gcloud components install gsutil docker-credentials-gcr` in the beginning and fail fast.

perotinus: That seems reasonable. Done.

hack/release.sh
```diff
@@ -29,6 +29,12 @@ trap clean_up EXIT
 BUILD_DATE="$(TZ=Etc/UTC date +%Y%m%d)"
 RELEASE_DIR="${TMPDIR}/${BUILD_DATE}"
 
+# Check for and install necessary dependencies.
+command -v bazel >/dev/null 2>&1 || { echo >&2 "Please install bazel before running this script."; exit 1; }
+command -v gcloud >/dev/null 2>&1 || { echo >&2 "Please install gcloud before running this script."; exit 1; }
```
madhusudancs: This is awesome! Thanks!

perotinus:
@font Prow.

hack/release.sh
perotinus: Thank you for all your feedback!

Issue #69, opened 2017-11-07
clusterregistry apiserver delegated vs standalone personality should be considered


font:
From [#8 (comment)](https://github.com/kubernetes/cluster-registry/issues/8#issuecomment-341819091):

How do we want to handle the different personalities that clusterregistry will assume when running as a delegated vs standalone apiserver?

Some things that will need to be addressed:

- Authn model: currently CR has built-in authn model. Will also need to delegate to core k8s apiserver. Need to allow both.
- Authz model: currently CR has always-allow authz model. Will also need to delegate to core k8s apiserver. Need to allow both.
- Necessary flags: handling common and mutually exclusive flags for standalone vs delegated
- Other stuff?

Some possible options:

- Add extra flag error checking. Possible confusion when user specifies combinations that are not compatible. This assumes we have several mutually exclusive options. It's not clear that we do yet but may in the future.
  - Roughly 8 for authn standalone and 8 for authn delegated. 
- Do we want to add new sub-commands to handle the different flags that could then trigger different paths for the authn/authz setup? This would seem strange for an apiserver. Are there any other apiservers doing this?
  - This could be something similar to hyperkube where the first argument specifies which "personality" to execute.
- Two separate binaries for delegated vs standalone. Probably try to avoid this unless we follow a hyperkube type model.
- Other options?

perotinus:
Given the changes in #120, @font and I believe that the best way to handle the two personalities is to have subcommands. So, `clusterregistry standalone [flags]` and `clusterregistry aggregated [flags]`. This works around the issue where the standalone and aggregated personalities have different authn/authz requirements, but with some flags with shared names, and should make it easier for someone to understand which flags apply to which usage of the tool. This is not necessarily something that other API servers are doing, but it does have some similarities with how `hyperkube` works (as @font mentioned above).

@pmorie @madhusudancs do you have any concerns with/thoughts about this approach? If not, we should start implementing it.

/assign @pmorie @madhusudancs

madhusudancs:
No concerns. +1 to splitting it into subcommands.

`hyperkube`'s case is very different though. `hyperkube` just serves as an anchor point for completely different binaries. But anyway, that distinction is not very relevant here though.

perotinus:
@font @pmorie I'm planning to take this on tomorrow if I don't hear any objections.

perotinus:
@font Ivan is going to take this on.

perotinus:
Fixed in #163.

PR #119, opened 2017-12-05
Cluster Registry support for delegated auth

font:
This is now ready to test.

Currently hitting re-defined flags being added due to standalone and delegated sharing similar flags e.g. `authentication-token-webhook-cache-ttl`, `client-ca-file`, etc. Need to either create a common interface for both with a common option flag structure they can both share, create different subcommands, or something else. Problem with a common interface and common option flag structure is that for delegated we rely on the `apiserver` API which we don't own, whereas for standalone we have our own implementation. Also see #69 for more details.

Fix for #8 when this PR is complete.

/sig multicluster

### DIFF
--- a/pkg/clusterregistry/options/options.go
+++ b/pkg/clusterregistry/options/options.go
@@ -28,29 +28,35 @@ import (
 
 // Runtime options for the clusterregistry-apiserver.
 type ServerRunOptions struct {
-	GenericServerRunOptions *genericoptions.ServerRunOptions
-	Etcd                    *genericoptions.EtcdOptions
-	SecureServing           *genericoptions.SecureServingOptions
-	Audit                   *genericoptions.AuditOptions
-	Features                *genericoptions.FeatureOptions
-	Authentication          *BuiltInAuthenticationOptions
+	GenericServerRunOptions  *genericoptions.ServerRunOptions
+	Etcd                     *genericoptions.EtcdOptions
+	SecureServing            *genericoptions.SecureServingOptions
+	Audit                    *genericoptions.AuditOptions
+	Features                 *genericoptions.FeatureOptions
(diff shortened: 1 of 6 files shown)


pkg/clusterregistry/options/options.go
```diff
@@ -28,29 +28,35 @@ import (
...
-		Audit:          genericoptions.NewAuditOptions(),
-		Features:       genericoptions.NewFeatureOptions(),
-		Authentication: NewBuiltInAuthenticationOptions().WithAll(),
+		Etcd:                     genericoptions.NewEtcdOptions(storagebackend.NewDefaultConfig("/registry/clusterregistry.kubernetes.io", nil)),
+		SecureServing:            genericoptions.NewSecureServingOptions(),
+		Audit:                    genericoptions.NewAuditOptions(),
+		Features:                 genericoptions.NewFeatureOptions(),
+		StandaloneAuthentication: NewBuiltInAuthenticationOptions().WithAll(),
+		DelegatedAuthentication:  genericoptions.NewDelegatingAuthenticationOptions(),
+		DelegatedAuthorization:   genericoptions.NewDelegatingAuthorizationOptions(),
```
pmorie: Instead of adding the options and then having to deal with the conflicts, I would suggest doing the following so that we can get aggregation working quickly:

- don’t add options to cli for now
- add only the boolean switch to enable aggregation
- when aggregation is turned on, just use the defaukt aggregation options

AFAICT, in the service-catalog helm chart, we do not actually set any of the auth-specific CLI flags and aggregation works just fine with the default options.

font:
@pmorie @perotinus This is now ready for review. We still need to figure out how best to handle the re-defined flags from standalone vs delegated auth.

font:
Closing in favor of #120 work done independently by @perotinus who is able to iterate on this during my trip to Kubecon.

PR #120, opened 2017-12-05
Implement delegated authn/z for the cluster registry

perotinus:
/sig multicluster

Addresses #8. Also see #119, which is parallel work done independently by @font.

Notable changes:

- Adds a flag to enable delegated authn/z for the cluster registry
- Adds a new role binding that allows the clusterregistry binary to access the `extension-apiserver-authentication` `ConfigMap`
- Remove non-DNS-compatible characters from the service account name
- Use the service account in the deployed Pods.


perotinus:
Tested and working with `crinit` in both standalone and aggregated modes.

perotinus:
/cc @font @pmorie @madhusudancs

This should be good to review now. I've made some changes to make it more like #119, but since @font is at Kubecon this week, I believe that I'll have more time to iterate on code changes.

pkg/clusterregistry/options/options.go
```diff
@@ -26,31 +26,38 @@ import (
...
+		Audit:                    genericoptions.NewAuditOptions(),
+		Features:                 genericoptions.NewFeatureOptions(),
+		StandaloneAuthentication: NewStandaloneAuthenticationOptions().WithAll(),
+		StandaloneAuthorization:  NewStandaloneAuthorizationOptions(),
+		DelegatingAuthentication: genericoptions.NewDelegatingAuthenticationOptions(),
+		DelegatingAuthorization:  genericoptions.NewDelegatingAuthorizationOptions(),
 
-		EventTTL: 1 * time.Hour,
+		EventTTL:         1 * time.Hour,
+		UseDelegatedAuth: false,
```
font: I feel like most may want to deploy the CR using the k8s aggregator and enable delegated auth by default. We can always change this in the future and if using `crinit` then it doesn't really matter too much.

pkg/clusterregistry/options/authorization.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/apiserver/pkg/authorization/authorizerfactory"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+)
+
+// StandaloneAuthorizerConfig configures an authorizer for the standalone
+// cluster registry.
+type StandaloneAuthorizerConfig struct {
+	AlwaysAllow bool
```
font: I suppose this structure and the `StandaloneAuthorizationOptions` structure will eventually diverge that separating them makes sense.

pkg/clusterregistry/options/authentication.go
```diff
@@ -149,17 +153,31 @@ func (s *BuiltInAuthenticationOptions) AddFlags(fs *pflag.FlagSet) {
...
+func (o *StandaloneAuthenticationOptions) ApplyTo(c *genericapiserver.Config) error {
+	var err error
+	if o.ClientCert != nil {
+		c, err = c.ApplyClientCert(o.ClientCert.ClientCA)
+		if err != nil {
+			return fmt.Errorf("unable to load client CA file: %v", err)
+		}
+	}
 
-		fs.DurationVar(&s.WebHook.CacheTTL, "authentication-token-webhook-cache-ttl", s.WebHook.CacheTTL,
```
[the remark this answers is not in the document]
perotinus: Hmm. That's a good point. At this point, I think that's acceptable, but it makes the solution of splitting `clusterregistry`'s personality into standalone and aggregated via subcommands more attractive and some solution necessary for a beta release (and ideally for an alpha).

Can you think of a good fix for this now? If this goes in without a fix, I'll file an issue to track making sure these flags are settable in standalone mode.

pkg/clusterregistry/options/authorization.go
perotinus: That was my thought as well. I considered removing this type, though, since it's not serving any particular purpose at the moment; I kept it for consistency with the other `Options` types and to support future divergence.

pkg/crinit/util/util.go
```diff
@@ -585,7 +585,7 @@ func createPVC(clientset client.Interface, namespace, svcName, etcdPVCapacity,
 
 func (o *SubcommandOptions) CreateAPIServer(cmdOut io.Writer, clientset client.Interface,
 	apiServerEnableHTTPBasicAuth, apiServerEnableTokenAuth bool, ips []string,
-	pvc *v1.PersistentVolumeClaim) error {
+	pvc *v1.PersistentVolumeClaim, serviceAccountName string, aggregated bool) error {
```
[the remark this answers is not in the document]
perotinus: Done.

pkg/crinit/util/util.go
```diff
@@ -618,7 +618,7 @@ func (o *SubcommandOptions) CreateAPIServer(cmdOut io.Writer, clientset client.I
 func createAPIServer(clientset client.Interface, namespace, name, serverImage,
 	etcdImage, advertiseAddress, credentialsName string, hasHTTPBasicAuthFile,
 	hasTokenAuthFile bool, argOverrides map[string]string,
-	pvc *v1.PersistentVolumeClaim, dryRun bool) (*appsv1beta1.Deployment, error) {
+	pvc *v1.PersistentVolumeClaim, serviceAccountName string, aggregated, dryRun bool) (*appsv1beta1.Deployment, error) {
```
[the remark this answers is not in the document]
perotinus: Done.

pkg/clusterregistry/options/options.go
perotinus: Good point. Done.

perotinus:
Thanks, @font! PTAL.

pkg/clusterregistry/options/authentication.go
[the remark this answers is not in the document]
font: Nope, I can't think of an easy way to do this. It seems like we could possibly implement our own `AddFlags()` that would add the common flags for both and then each of the unique APIs can add the rest. We can come back and fix this.

pkg/clusterregistry/options/authorization.go
font: sgtm

pkg/clusterregistry/options/options.go
font: ?

perotinus: Hmm. I meant to change this to `true`, let me fix that now.

perotinus:
@font OK, I must have forgotten to push last time. I pushed a comment with your comments addressed.

perotinus:
Also, thanks for the review!

perotinus:
FYI, I'll hold off on merging this until tomorrow in case anyone else wants to take a look at it.

pkg/clusterregistry/options/options.go
```diff
@@ -26,31 +26,38 @@ import (
...
+		Audit:                    genericoptions.NewAuditOptions(),
+		Features:                 genericoptions.NewFeatureOptions(),
+		StandaloneAuthentication: NewStandaloneAuthenticationOptions().WithAll(),
+		StandaloneAuthorization:  NewStandaloneAuthorizationOptions(),
+		DelegatingAuthentication: genericoptions.NewDelegatingAuthenticationOptions(),
+		DelegatingAuthorization:  genericoptions.NewDelegatingAuthorizationOptions(),
 
-		EventTTL: 1 * time.Hour,
+		EventTTL:         1 * time.Hour,
+		UseDelegatedAuth: true,
```
madhusudancs: Do we want `true` to be the default? Is this consistent with the behavior elsewhere? I mean, do we default to delegated+aggregated mode in other places too?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -186,18 +192,18 @@ func Run(opts *aggregatedClusterRegistryOptions, cmdOut io.Writer,
 		return err
 	}
 
-	err = opts.CreateAPIServer(cmdOut, hostClientset, false, false, ips, pvc)
+	sa, err := createRBACObjects(cmdOut, hostClientset, opts)
```
madhusudancs: Why is this an odd man out? Why does this take opts as an argument instead of taking it as a receiver like other method calls around?

pkg/clusterregistry/options/options.go
perotinus: @font suggested that I default this to true, because we expect that most people will want to deploy an aggregated registry. In the end, if people use `crinit`, this flag will not be exposed to them, so I think we have some flexibility to change it later.

pkg/crinit/aggregated/aggregated.go
perotinus: Hmm. I may have just missed this in review. @font?

I think we intend to refactor this anyway, so that opts is not the receiver, so I'm not so concerned about having this be inconsistent for a bit.

pmorie:
LGTM - I’m happy to merge this unless @perotinus wants to do anything else - will merge in a bit unless there are objections.

pkg/crinit/aggregated/aggregated.go
pmorie: We can fix this in a follow-up.  Remember, this isn't even alpha yet :)

font: No particular reason other than this function is in this package whereas the others are in the `util` package which contains the defined `SubcommandOptions` structure. This `opts` gets passed as the `aggregatedClusterRegistryOptions` type in this package whereas the other methods in `util` are invoked as `opts.SubcommandOptions` (the embedded type) receivers.

Also as @perotinus states, this is going to be refactored further as there is an open PR #127 for it that may already take care of it (haven't reviewed it yet).

pkg/clusterregistry/options/options.go
madhusudancs: Ok.

pkg/crinit/aggregated/aggregated.go
madhusudancs: @perotinus @pmorie @font fair enough. Good to know that you are all aware of it and there is an open PR to refactor it.

Issue #22, opened 2017-10-11
Add validation logic to ensure that metadata.ClusterName is not set


perotinus:
cf [this comment](https://docs.google.com/a/google.com/document/d/1Oi9EO3Jwtp69obakl-9YpLkP764GZzsz95XJlX1a960/edit?disco=AAAABUIGW-g) in the cluster registry API doc.

PR #123, opened 2017-12-06
Add validation that metadata.clusterName is not set.

perotinus:
/sig multicluster

Addresses #22

### DIFF
--- a/pkg/apis/clusterregistry/validation/validation.go
+++ b/pkg/apis/clusterregistry/validation/validation.go
@@ -23,14 +23,24 @@ import (
 )
 
 func ValidateCluster(cluster *clusterregistry.Cluster) field.ErrorList {
-	return validation.ValidateObjectMeta(&cluster.ObjectMeta, false, validation.ValidateClusterName, field.NewPath("metadata"))
+	allErrs := validation.ValidateObjectMeta(&cluster.ObjectMeta, false, validation.ValidateClusterName, field.NewPath("metadata"))
+	allErrs = append(allErrs, validateClusterName(cluster)...)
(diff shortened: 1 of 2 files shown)


pkg/apis/clusterregistry/validation/validation.go
```diff
@@ -32,5 +34,13 @@ func ValidateClusterUpdate(cluster, oldCluster *clusterregistry.Cluster) field.E
 		allErrs = append(allErrs, field.Invalid(field.NewPath("meta", "name"),
 			cluster.Name+" != "+oldCluster.Name, "cannot change cluster name"))
 	}
+	allErrs = append(allErrs, validateClusterName(cluster)...)
 	return allErrs
 }
+
+func validateClusterName(cluster *clusterregistry.Cluster) field.ErrorList {
+	if len(cluster.ClusterName) > 0 {
+		return field.ErrorList{field.Invalid(field.NewPath("meta", "clusterName"), "len(ClusterName) > 0", "clusterName is not used and must not be set.")}
```
madhusudancs: Isn't `field.NewPath("meta" ...)` actually `field.NewPath("metadata" ...)`? I see this is also repeated elsewhere in this file.

perotinus: Hmm. I was wondering that myself, but just assumed that it was correct. I just searched through the Kubernetes codebase, and found that this is usually `"metadata"`, though, so I've changed it here.

madhusudancs: Makes sense. Thanks for fixing this.

Issue #110, opened 2017-11-30
Get versioning info for the built tools from the environment


perotinus:
/sig multicluster

Kubernetes uses bazel `--workspace_status` to generate version info and stamp the binaries. We should do the same thing.

e.g.,
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/BUILD#L20
uses
https://github.com/kubernetes/kubernetes/blob/master/pkg/version/def.bzl
which is populated by
https://github.com/kubernetes/kubernetes/blob/master/hack/print-workspace-status.sh
run because of
https://github.com/kubernetes/kubernetes/blob/master/.bazelrc
which uses info genereated by
https://github.com/kubernetes/kubernetes/blob/master/hack/lib/version.sh

PR #124, opened 2017-12-06
Stamp bazel-built binaries with version informaton

perotinus:
/sig multicluster

We can determine a better versioning/tagging strategy later. This is mostly pulled from k/k, with some simplifications. At the least, it gives us a way to track what versions of the tools people are using, which is a big win.

/cc @pmorie @madhusudancs @font

### DIFF
--- a/hack/print-workspace-status.sh
+++ b/hack/print-workspace-status.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This command is used by bazel as the workspace_status_command
+# to implement build stamping with git information. It pulls information
+# from tags named like vX.Y.Z with an optional postfix, e.g., v1.2.3-alpha.
+
+set -euo pipefail
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+GIT=(git --work-tree "${REPO_ROOT}")
+
+if GIT_COMMIT=$("${GIT[@]}" rev-parse "HEAD^{commit}" 2>/dev/null); then
+  # Check if the tree is dirty. Default to dirty
+  GIT_TREE_STATE="dirty"
+  if git_status=$("${git[@]}" status --porcelain 2>/dev/null) && [[ -z ${git_status} ]]; then
+    GIT_TREE_STATE="clean"
+  fi
+
+  # Use git describe to find the version based on annotated tags.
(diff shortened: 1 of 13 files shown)


perotinus:
Example:

```sh
$ bazel-bin/cmd/clusterregistry/clusterregistry --version
version.Info{Major:"", Minor:"", GitVersion:"v0.0.1-rc.0.1+02e069d270221c-dirty", GitCommit:"02e069d270221c238266b834a00fcecf9e86e051", GitTreeState:"dirty", BuildDate:"2017-12-06T03:36:38Z", GoVersion:"go1.9.1", Compiler:"gc", Platform:"linux/amd64"}
$
```

perotinus:
Addresses #110

hack/print-workspace-status.sh
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+
+# This command is used by bazel as the workspace_status_command
+# to implement build stamping with git information. It pulls information
+# from tags named like vX.Y.Z with an optional postfix, e.g., v1.2.3-alpha.
+
+set -euo pipefail
+
+SCRIPT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT=$(cd ${SCRIPT_ROOT}/..; pwd)
```
madhusudancs: Use `git rev-parse --show-toplevel` instead?

pkg/version/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
```
madhusudancs: Could this all be just vendored from k/k and re-used? By this I mean everything in this directory.

hack/print-workspace-status.sh
perotinus: Hmm. That doesn't work if the script is being run from outside of the repository. I'm not sure that's a major concern since bazel is running it, and for bazel to run it must be in the workspace.

Thoughts?

pkg/version/BUILD.bazel
perotinus: Potentially, but I don't think it would be easy. There are some modifications that I've made to the files to better support the cluster registry:

- changing the package in `def.bzl`
- simplifying the default variable definitions in `base.go`
- not using the `gitMajor` and `gitMinor` flags, which are said to be deprecated

I also don't particularly want to tie the cluster registry to the way that Kubernetes does its versioning, and would rather have room to diverge as needed. Also, vendoring in k/k will potentially make cloning/updating deps/etc more cumbersome, and I'd rather avoid having the cluster registry depend on k/k directly.

If you think there's value in a consistent versioning model across the k8s projects, I can file (or find and +1) an issue that tracks extracting this version functionality into a shared location. I only need the bazel-based process, though, and I expect there would be a lot of k/k-specific stuff that would be hard to usefully extract.

hack/print-workspace-status.sh
madhusudancs: That's a fair point. I am not hell bent on making this change. I just thought `rev-parse` was more readable and closer to what you were trying to achieve here. But whatever works.

pkg/version/BUILD.bazel
madhusudancs: What you are saying makes sense. I don't think we should make versioning consistent across k8s projects, but just share the infrastructure if they are exactly identical. If you are planning to diverge, then keep them separate makes sense.

hack/print-workspace-status.sh
perotinus: I tried this out with bazel just to see, and it works fine, so I'm going to make this simplification. I like having fewer variables, and the `SCRIPT_ROOT` command always feels a bit voodoo-ish to me.

Issue #23, opened 2017-10-12
Set up per-PR tests and other CI


perotinus:
Once #19 goes in, there will be an incomplete but runnable suite of bazel tests that should be run before PR submission.

This involves adding some configuration to the Prow config (https://github.com/kubernetes/test-infra/blob/master/prow/config.yaml) and the job config (https://github.com/kubernetes/test-infra/blob/master/jobs/config.json).

Since it is currently necessary to run `./update-codegen.sh` and `bazel run //:gazelle` before running the bazel tests, I expect that we will have to use a custom execution scenario (https://github.com/kubernetes/test-infra/blob/master/scenarios/execute.py) and call the bazel scenario directly after running the codegen commands.

perotinus:
Some specific tasks:
- [] Run the bazel tests on a per-PR basis
- [] Run standard validation checks on a per-PR basis (ideally from https://github.com/kubernetes/repo-infra/tree/master/verify)
- [] Add a CI job that runs the bazel tests
- [] Rearrange the sig-multicluster dashboard on testgrid into federation and cluster-registry sections, and expose the cluster-registry CI jobs there

pmorie:
@perotinus I believe in the comment above, the CI job that runs the bazel tests is complete - accurate?

perotinus:
I was thinking of CI as a continuously-running job rather than a per-PR job. At this point, I don't think there's a practical need for CI, since there are no external dependencies in tests, and the tests run by the CI would be exactly the same as those run by the per-PR job, but it may be worth setting up so that, in the future, there is a place to put tests that should not block PRs but are worth running consistently.

font:
@perotinus @pmorie @madhusudancs Should we add one job that runs all the verify steps or add one job for each verify step? Need to consider the pros and cons to each.

One job for each allows the testgrid to show each one separately and allows one to see results for each separately. This could be useful if there are multiple failures.

One job for all would probably run faster due to reduced time for scheduling the job. We'd also have to have a script that will run each one separately and report the combined results at the end so that it doesn't exit on the first failure. This would help triage and fix multiple failures faster, otherwise one would have to iterate several times for each failed step.

madhusudancs:
> But we miss out on parallelism of having multiple jobs. We'd also have to have a script that will run each one separately and report the combined results at the end so that it doesn't exit on the first failure.

We can configure these the way we want in a single job case too right? Run the verify steps in parallel in a single job and aggregate results in the end as you propose?

[...]

PR #135, opened 2017-12-12
Update bazel BUILD files

font:
/sig multicluster

### DIFF
--- a/pkg/clusterregistry/BUILD.bazel
+++ b/pkg/clusterregistry/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "cluster.go",
+        "doc.go",
         "server.go",
     ],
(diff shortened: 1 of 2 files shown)


pmorie:
Seems like we need a bazel build verification step.

font:
@pmorie There is a bazel build that is run as part of testing PRs, it just needs the `ok-to-test` label added. These missing `doc.go` files did not cause build failures.

perotinus:
There is a script in `hack`, `hack/verify-bazel.sh`, but this hasn't been set up to run on Prow yet. It should not be that hard to add a pull job that runs all the verify scripts (and potentially some more that run things like `gofmt` and `go lint`).

pmorie:
@perotinus Seems like the verify stuff is covered by https://github.com/kubernetes/cluster-registry/issues/23, agree?

perotinus:
@pmorie Yes, agreed.

PR #136, opened 2017-12-12
Move CRUD test to test/integration dir

font:
/sig multicluster


font:
This only ran 5 of the current 6 tests because it gates this PR:
https://github.com/kubernetes/test-infra/pull/5912

pmorie:
Sounds like it won't actually work until this PR is merged?  That seems not-optimal.

pmorie:
Sounds to me like we can't know that this test actually runs correctly in the new location, because the CI won't run it until the PR is merged so we can merge the prow PR - is that accurate?

font:
@pmorie Yes that's correct. It seems like ideally we'd want prow to run a test script that is provided by our repo so that any changes to it are completely confined to our repo. We actually used to have a script `./hack/bazel-test.sh` that contained the `bazel test` command and prow would execute the script. If that were still the case, this PR would also be updating that script's test command and no changes to `test-infra` would be required.

I'm not sure why `./hack/bazel-test.sh` was removed. But I see all the other jobs in prow are doing the same thing i.e. providing the `bazel test` command directly in their job config as part of a `kubernetes_bazel` scenario.

@perotinus @madhusudancs Should we revert to use a script provided by our repo?

perotinus:
@font I don't think so: all `hack/bazel-test.sh` was doing was running the `bazel test` command that's in the development docs, so running it locally in the repo should provide an equivalent result (modulo prow issues that running the script by-hand may or may not find).

An alternative strategy that we could try is changing the `--test` flag to something like `//... -//vendor/... -//cmd/clusterregistry:push-clusterregistry-image`

font:
@perotinus That should work for now. I've updated the `test-infra` PR. Thanks!

perotinus:
Tests are passing, and including the //test/... target, so I'm going to merge this.

BenTheElder:
Side note: in test-infra we've moved to rm'ing tests from `vendor/` since dep is already moving towards this. Then there are no bazel test targets to exclude. `//cmd/clusterregistry:push-clusterregistry-image` shouldn't match as a test target (?)

perotinus:
@BenTheElder The problem that I remember running into is that `bazel`'s analysis phase chokes on the `//cmd/clusterregistry:push-clusterregistry-image` target because it requires a `define`, which caused the initial analysis to fail before it even started running the tests. I think there may have been a similar issue in `vendor`, but I'm not sure.

PR #139, opened 2017-12-13
Add init subcommands to the aggregated and standalone crinit commands.

perotinus:
/sig multicluster

Also adds a `--short` flag for the version.

### DIFF
--- a/pkg/crinit/aggregated/aggregated.go
+++ b/pkg/crinit/aggregated/aggregated.go
@@ -42,17 +42,17 @@ import (
 )
 
 var (
-	longCommandDescription = `
-	Aggregated initializes an aggregated cluster registry.
+	longInitCommandDescription = `
+	Init initializes an aggregated cluster registry.
 
 	The aggregated cluster registry is hosted inside a Kubernetes
 	cluster and registers its API with the Kubernetes API aggregator.
 	The host cluster must be specified using the --host-cluster-context flag.`
-	commandExample = `
+	initCommandExample = `
 	# Initialize an aggregated cluster registry named foo
 	# in the host cluster whose local kubeconfig
 	# context is bar.
-	crinit aggregated foo --host-cluster-context=bar`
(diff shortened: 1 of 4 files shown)


pkg/crinit/aggregated/aggregated.go
```diff
@@ -113,10 +113,16 @@ func NewCmdAggregated(cmdOut io.Writer, pathOptions *clientcmd.PathOptions, defa
...
-		Short:   "Initializes an aggregated cluster registry",
-		Long:    longCommandDescription,
-		Example: commandExample,
+		Use:   "aggregated",
+		Short: "Subcommands to manage an aggregated cluster registry",
+		Long:  "Commands used to manage an aggregated cluster registry. That is, a cluster registry that is aggregated with another Kubernetes API server.",
+	}
+
+	initCmd := &cobra.Command{
+		Use:     "init",
```
madhusudancs: Shouldn't this be `"init  CLUSTER_REGISTRY_NAME --host-cluster-context=HOST_CONTEXT"`?

pkg/crinit/standalone/standalone.go
```diff
@@ -70,10 +70,16 @@ func NewCmdStandalone(cmdOut io.Writer, pathOptions *clientcmd.PathOptions, defa
 	opts := &standaloneClusterRegistryOptions{}
 
 	cmd := &cobra.Command{
-		Use:     "standalone CLUSTER_REGISTRY_NAME --host-cluster-context=HOST_CONTEXT",
+		Use:   "standalone <subcommand>",
```
madhusudancs: Add `<subcommand>` to the aggregated version too or remove it from here too. Either way is fine, but being consistent would be better.

perotinus: Good point. Oddly, the `<subcommand>` bit didn't appear when I was running the command, so I'll remove it and let the `cobra` machinery handle describing the behavior.

PR #140, opened 2017-12-14
Add cloudProvider to example cluster configs

font:
/sig multicluster


pmorie:
/ok-to-test

my bad, i keep forgetting that i need to do this on your PRs, @font

perotinus:
One point of interest: it might be worthwhile to write a test that this YAML file can be added to a cluster registry without errors. This could perhaps be added to the integration test.

PR #141, opened 2017-12-14
Update OpenAPI generated code and spec

font:
We'll want to add the `hack/verify-codegen.sh` and `hack/verify-openapi-spec.sh` scripts as a prow job.

/sig multicluster


pmorie:
thanks, @font - think you can tackle the prow job?

font:
@pmorie Yep, I planned to but need to find the best practice for https://github.com/kubernetes/cluster-registry/issues/23#issuecomment-351601031. After talking with sig-testing, they don't seem to have an actual preference yet.

PR #145, opened 2017-12-15
Add script to verify all generated source for prow job

font:
Will turn this into a single prow job.

/sig multicluster


font:
Not ready for this yet. May go a different route.

PR #151, opened 2017-12-18
Use the generated version in the clusterregistry API server version field.

perotinus:
/sig multicluster


perotinus:

One interesting quirk about this: since the OpenAPI spec version is set from the tag version, this file cannot be updated in the tagged release: it has to be in a follow-up commit that comes just after the release. I don't think this is a huge deal, but it's worth noting.

PR #153, opened 2017-12-19
Install curl for use in prow job's k8s test image

font:
/sig multicluster


perotinus:
/lgtm

Looks like this fixes the issue with generating the swagger spec from the results of the Prow job.

font:
Updated with a fix to specify GOROOT so that the bazel Go toolchain is used. Thanks @perotinus for identifying the fix!

Issue #159, opened 2017-12-21
Fix OpenAPI code generation for MicroTime and Time structures


font:
Our OpenAPI code generation is not properly detecting the `MicroTime` and `Time` custom OpenAPI schema defined in their respective files. For example, `MicroTime` defines https://github.com/kubernetes/apimachinery/blob/20bbfef868144faf29af69ddb2f01646ead5c1a1/pkg/apis/meta/v1/micro_time.go#L152-L161 which should be picked up by the `openapi-gen` code generator to generate the following diff output: 

```diff
diff --git a/pkg/apis/clusterregistry/v1alpha1/zz_generated.openapi.go b/pkg/apis/clusterregistry/v1alpha1/zz_generated.openapi.go
index 01aa6f2e..a5c5a8c2 100644
--- a/pkg/apis/clusterregistry/v1alpha1/zz_generated.openapi.go
+++ b/pkg/apis/clusterregistry/v1alpha1/zz_generated.openapi.go
@@ -22,6 +22,7 @@ package v1alpha1
 
 import (
 	spec "github.com/go-openapi/spec"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	common "k8s.io/kube-openapi/pkg/common"
 )
 
@@ -941,23 +942,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 			},
 			Dependencies: []string{},
 		},
-		"k8s.io/apimachinery/pkg/apis/meta/v1.MicroTime": {
-			Schema: spec.Schema{
-				SchemaProps: spec.SchemaProps{
-					Description: "MicroTime is version of Time with microsecond level precision.",
-					Properties: map[string]spec.Schema{
-						"Time": {
-							SchemaProps: spec.SchemaProps{
-								Type:   []string{"string"},
-								Format: "date-time",
-							},
-						},
-					},
-					Required: []string{"Time"},
-				},
-			},
-			Dependencies: []string{},
-		},
+		"k8s.io/apimachinery/pkg/apis/meta/v1.MicroTime": v1.MicroTime{}.OpenAPIDefinition(),
 		"k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta": {
 			Schema: spec.Schema{
 				SchemaProps: spec.SchemaProps{
@@ -1400,23 +1385,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
[...]

PR #160, opened 2017-12-21
Fix for #159

font:
Fix for #159.

Thanks to @luksa for identifying a dependency issue!

/sig multicluster


perotinus:
/approve

Thanks @font!

PR #155, opened 2017-12-19
Implement standalone webhook authoritzation.

perotinus:
/sig multicluster

Also adds an example using a webhook for authorization.

Fixes #46.

/cc @font @pmorie @madhusudancs

PR #165, opened 2018-01-04
Fixes for webhook authorization

perotinus:
/sig multicluster

Follow-up to comments #155, along with a fix that should have gone into that PR.

/cc @madhusudancs @font @pmorie

### DIFF
--- a/examples/webhook-authz/README.md
+++ b/examples/webhook-authz/README.md
@@ -21,12 +21,15 @@ production environment.
 openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -nodes
 ```
 
-4.  Create a ConfigMap that contains the files needed for the cluster registry
-    webhook authorizer. If you deployed your cluster registry into a namespace
-    other than `clusterregistry`, change the namespace below accordingly.
+4.  Create a Secret that contains the files needed for the cluster registry
+    webhook authorizer (i.e., the `rules.hcl` file, the configuration file that
+    configures the cluster registry API server to use the webhook authorizer,
+    and the SSL key and certificate). If you deployed your cluster registry into
+    a namespace other than `clusterregistry`, change the namespace below
+    accordingly.
 
 ```sh
-kubectl create configmap bitesize -n clusterregistry \
+kubectl create secret generic bitesize -n clusterregistry \
   --from-file=./rules.hcl --from-file=./webhook-config.yaml \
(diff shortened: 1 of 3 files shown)


examples/webhook-authz/README.md
```diff
@@ -52,17 +55,17 @@ kubectl apply -f ./authz.yaml -n clusterregistry
...
       volumes:
       ... existing volumes ...
       - name: bitesize
-        configMap:
-          name: bitesize
+        secret:
+          secretName: bitesize
       - name: ca-certificates
         hostPath:
           path: /etc/ssl/certs
```
font: Do you know why this is needed?

examples/webhook-authz/README.md
```diff
@@ -100,7 +103,7 @@ or create/update them as `testuser`:
 
 ```
 sh
```
font: Combine with previous line.

examples/webhook-authz/README.md
```diff
@@ -100,7 +103,7 @@ or create/update them as `testuser`:
...
-$ kubectl apply -f cluster.yaml --context <your_cluster_registry_context> --as testuser
+$ kubectl apply -f ../samplecontainer/cluster.yaml --context <your_cluster_registry_context> --as testuser
 Error from server (Forbidden): error when applying patch:
```
font: The error I received was a `create` error and not a `patch` error since it was the first time created:
```
Error from server (Forbidden): error when creating "../samplecontainer/cluster.yaml": clusters.clusterregistry.k8s.io is forbidden: User "testuser" cannot create clusters.clusterregistry.k8s.io at the cluster scope: Not allowed
```

perotinus: Good point. Updated.

examples/webhook-authz/README.md
perotinus: Done.

examples/webhook-authz/README.md
perotinus: I ran into a problem where the authorizer and/or clusterregistry didn't work unless I added a volume that pointed to the host's `/etc/ssl/certs`. I just updated my local deployment to not have this volume mounted and things still worked fine, so I'm wondering if this is an issue depending on how you create the cert that is used to secure communication between the webhook and the cluster registry API server. Perhaps I was using a cert that relied on one of the system certs?

It doesn't seem strictly necessary to making this run, so I think I'll remove it.

PR #174, opened 2018-01-13
Update and simplify Go source file verification scripts

font:
- Update OpenAPI spec after cluster registry v0.0.2 released.

- Updated the script that runs all Go source file verify scripts to not take any arguments. This helps simplify it, easier to document it, and easier to run as a Prow job.

- Also updated individual verify scripts to allow being run from any directory.

- Fixed the `gofmt` errors.

/sig multicluster

### DIFF
--- a/api/swagger.json
+++ b/api/swagger.json
@@ -2,7 +2,7 @@
   "swagger": "2.0",
   "info": {
    "title": "Cluster Registry",
-   "version": "v0.0.1"
+   "version": "v0.0.2"
   },
   "paths": {
    "/apis/": {
--- a/hack/verify-go-src.sh
+++ b/hack/verify-go-src.sh
@@ -13,82 +13,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This script runs all of the Go source verification scripts inside of the
+# ./hack/go-tools directory. The success or failure of each script is outputted
+# in green or red colored text, respectively. If any script fails, an error is
+# returned, otherwise returns 0.
+
 set -euo pipefail
 
-# This script is intended to be used via subtree in a top-level directory:
-# <repo>/
-#  repo-infra/
-#    verify/
-# Or via vendoring and passing root directory as vendor/repo-infra/verify-*.sh --rootdir **full path to your repo dir**
-# <repo>/
-#   vendor/
-#      repo-infra/
-#         ...
-# 
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_ROOT}/.." && pwd)"
+TMP_GOPATH="$(mktemp -d /tmp/gopathXXXXXXXX)"
+TMP_REPO_ROOT="${TMP_GOPATH}/src/k8s.io/cluster-registry"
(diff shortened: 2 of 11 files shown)


api/swagger.json
```diff
@@ -2,7 +2,7 @@
   "swagger": "2.0",
   "info": {
    "title": "Cluster Registry",
-   "version": "v0.0.1"
```
perotinus: Can you make this commit first in the commit tree? I think it would be cleaner if this is the first thing done after the release.

perotinus: Also, my apologies for forgetting to do this last week!

hack/verify-go-src.sh
```diff
@@ -99,10 +48,22 @@ function run-checks {
   done
 }
 
-if ${SILENT} ; then
-  echo "Running in silent mode, run with -v if you want to see script logs."
+# Set up the temporary GOPATH
+mkdir -p "${TMP_REPO_ROOT}"
+cp -r "${REPO_ROOT}/"* "${TMP_REPO_ROOT}"
+export GOPATH="${TMP_GOPATH}"
```
perotinus: Why the temporary `GOPATH`?

hack/verify-go-src.sh
```diff
@@ -15,80 +15,29 @@
 
 set -euo pipefail
```
perotinus: Can you add some documentation describing what this script is doing? It's not immediately obvious IMO from the code, and I think a quick summary will make maintenance easier.

api/swagger.json
font: I know it's not apparent, but it actually already is as I made sure to do that for the same reasons you suggest. Strangely enough, the commits aren't ordered correctly in the commit history through github's web UI. But if you click to `Browse the repository at this point in the history` on any or all commits (last commit is `Fix gofmt errors`), then click on commits, you'll see the history shows it's the first commit. Perhaps this is because of the tide pool error.

hack/verify-go-src.sh
font: Will do.

hack/verify-go-src.sh
font: This helps to run this script in a vanilla Prow environment because some of the scripts called by this one e.g. `verify-govet.sh`, run `go list` which will print the import path for each package (e.g. `k8s.io/cluster-registry/...`) to pass into `go vet` and `go` will not find it correctly in `GOPATH` unless the package exists using that same import path. So this temporary `GOPATH` is set up to replicate the same import path location.

I will add a comment to this effect.

api/swagger.json
perotinus: Huh. Oh well, thanks for being on top of things. :)

perotinus:
@font Did you upload your latest commits?

font:
@perotinus I hadn't, but I've now uploaded them.

hack/verify-go-src.sh
```diff
@@ -99,10 +52,27 @@ function run-checks {
   done
 }
 
-if ${SILENT} ; then
-  echo "Running in silent mode, run with -v if you want to see script logs."
+# Set up the temporary GOPATH. This helps to run this script in a vanilla
+# environment e.g. Prow, because the verify-govet.sh script called by this one
```
perotinus: nit: "... environment, e.g. for Prow, because ..."

font:
@perotinus Updated.

perotinus:
/lgtm

Thanks!

Issue #172, opened 2018-01-11
Get current clusters failed due to User `apiserver-client` permission


patrickshan:
/sig multicluster

**What happened**:
After provisioning cluster registry using crinit (actually it hangs at the last step to wait api server to come up but the actually pods and service are UP), when you do a get cluster operation, it will fail:
`
$ kubectl get clusters
Error from server (Forbidden): clusters.clusterregistry.k8s.io is forbidden: User "apiserver-client" cannot list clusters.clusterregistry.k8s.io at the cluster scope
`

**What you expected to happen**:
It should return "No resources found"

**How to reproduce it (as minimally and precisely as possible)**:

1. use crinit to provision cluster registry inside an existing kube cluster:
```
./crinit aggregated init lab-cluster-registry --etcd-pv-storage-class ebs-volume --image 'gcr.io/crreleases/clusterregistry:latest_nightly' --host-cluster-context cluster-0.ap-southeast-2.lab
```
2. the above command will hang at the last step `Waiting for the cluster registry API server to come up...` when you check the pods, etcd container is UP while clusterregistry container complains about certificates error
3. Update secret object `lab-cluster-registry-apiserver-credentials` to use cluster's existing CA to replace `ca.crt`
4. kill the existing pods to pick up the new CA and the pod will be UP and running
```
$ kubectl logs -f lab-cluster-registry-apiserver-847849b6b5-msl66 clusterregistry
I0111 04:10:17.060912       1 serve.go:85] Serving securely on 0.0.0.0:8443
```
5. run kubectl get clusters which will get the error above
```
$ kubectl get clusters
Error from server (Forbidden): clusters.clusterregistry.k8s.io is forbidden: User "apiserver-client" cannot list clusters.clusterregistry.k8s.io at the cluster scope
```

**Anything else we need to know?**:
I also tried to update clusterregistry.k8s.io:apiserver clusterrolebindings to include `apiserver-client` User but that didn't solve the problem:
```
$ kubectl get clusterrolebindings clusterregistry.k8s.io:apiserver -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: 2018-01-11T04:09:18Z
  labels:
    app: clusterregistry
  name: clusterregistry.k8s.io:apiserver
  resourceVersion: "1034807"
[...]

PR #175, opened 2018-01-16
Add link to userguide to enable k8s aggregation layer

font:
/sig multicluster

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -53,7 +53,13 @@ $
 
 ### Aggregated API server
 
-You can deploy the cluster registry as an aggregated API server like so:
+Before deploying the cluster registry as an aggregated API server, take a look at
+https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#enable-apiserver-flags
+which talks about what `kube-apiserver` flags need to be enabled in order to
+enable the aggregation layer.
+
+Once the `kube-apiserver` aggregation layer is enabled, you can deploy the
+cluster registry as an aggregated API server like so:
 
 ```sh
 ./crinit aggregated init <cluster_registry_context> --host-cluster-context=<your_cluster_context>


perotinus:
@font I'll LGTM this after #174 goes in. Thanks, this is a good change!

docs/userguide.md
```diff
@@ -53,7 +53,13 @@ $
...
 ### Aggregated API server
 
-You can deploy the cluster registry as an aggregated API server like so:
+Before deploying the cluster registry as an aggregated API server, take a look at
+https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#enable-apiserver-flags
+which talks about what `kube-apiserver` flags need to be enabled in order to
+enable the aggregation layer.
+
+Once the `kube-apiserver` aggregation layer is enabled, you can deploy the
+cluster registry as an aggregated API server like so:
```
perotinus: Actually, looking at the context in #172, maybe we should add a section about the names specifically? Or else find a link to a site that explains it in more detail? The link here gives a brief overview of the flags, but doesn't necessarily explain how they should be used with a custom CA.

font: Unfortunately, there is lack of proper documentation on this topic. The link to the list of flags doesn't have much detail on how values of those flags can vary in different configurations, but AFAICT it's all we have.

I also thought about documenting it some myself, but I'm not 100% certain what all the security ramifications are for specifying multiple or removing the `--requestheader-allowed-names` flag. So I figured I should avoid providing any potentially misleading information in our docs that others may attempt to use in a production deployment of the CR. Especially when the k8s docs themselves do not document it.

Thoughts?

perotinus: That seems reasonable to me: if the k8s docs don't discuss this, it's better for us not to provide info.

We can revisit this, though, if someone else brings up the same issue.

PR #176, opened 2018-01-16
Add type info to the AuthInfo object.

### LINKED ISSUE #169: 'Type' field for the AuthInfo object
/sig multicluster

@pmorie suggested that the `AuthInfo` object could also have a `Type` field at the top level. This would help clients to distinguish what they should do with the contents of the `AuthInfo` object.

I think this is a good thing to add: every `AuthInfo` object does require type information in order to be useful, and pulling it into a field allows for potential validation/development a _de facto_ or formalized standard of types.

I'd suggest considering having the `Type` field be an object (`TypeInfo`) to allow future extensions to the type information. For example, it would the be possible to express different versions of an authentication type without having to crush them into `X-v1`, `X-v2` strings, and so on; or, to have subtypes that are specialized varieties of one type.

perotinus:
/sig multicluster

Fixes #169.

/cc @pmorie @madhusudancs @font 
Please take a look at this; this is an API change and I want to make sure that we're happy with it, or that we have a chance to suggest more fields that should be in `AuthProviderType`.


madhusudancs:
Looks reasonable to me. Nothing controversial here.

pkg/apis/clusterregistry/types.go
```diff
@@ -114,12 +114,31 @@ type AuthProviderConfig struct {
...
+	// +optional
+	Type AuthProviderType
+
 	// Config is a map of values that contains the information necessary for a
 	// client to determine how to authenticate to a Kubernetes API server.
 	// +optional
 	Config map[string]string
 }
 
+// AuthProviderType contains metadata about the auth provider. It should be used
```
font: nit: This says metadata but it just contains the name right now.

pkg/apis/clusterregistry/types.go
```diff
@@ -114,12 +114,31 @@ type AuthProviderConfig struct {
...
+// by clients to differentiate between different kinds of auth providers, and so
```
font: nit: `s/so//`?

pkg/apis/clusterregistry/types.go
```diff
@@ -114,12 +114,31 @@ type AuthProviderConfig struct {
...
+// select a relevant provider for the client's configuration. For example, a
+// controller would look for a provider type that denotes a service account
+// that it should use to access the cluster, whereas a user would look for a
+// provider type that denotes an authentication system from which they should
+// request a token.
+type AuthProviderType struct {
+	// Name is the name of the auth provider.
+	// +optional
+	Name string
```
font: If we don't add anything else to this, it seems somewhat similar to the `AuthProviderConfig:Name` field. That is, couldn't one use the `AuthProviderConfig:Name` field to designate the auth provider's name? The example of a controller vs a user could still use the `AuthProviderConfig:Name` to find the relevant auth config.

Is this `AuthProviderType` addition still relevant if we consider making cluster namespace scoped?

perotinus: Yes, I think so. The idea of the `Type` field is that it specifically allows for repeated `AuthProviders` of one type with different info, and makes this more obvious than having that info in the bag of data. For example, imagine that you have an AuthProvider type that is "GCP service account". You can now have two different AuthProvider instances with that type, but with different names and representing different service accounts.

Also, my intention here is that this may expand in the future in a backwards-compatible way. For example, I could see adding a version field. I'd also rather people not try to squeeze this information into the name (e.g., gcp-service-account_foo, gcp-service-account-bar for two different AuthProviders). I think this field makes writing a plugin a bit easier as well, and provides a standard way to encode this. It's something that every `AuthProvider` object will need to do somehow, so making it top-level and typed seems like a reasonable thing to do.

font: Thanks! I was interested to hear your thoughts on this. I agree, I had thought that reusing an auth provider type in the top level config name would require overloading that name field quite a bit. I also think that explicitly specifying this `AuthProviderType` struct field makes it prescriptive so that users don't try to embed that sort of metadata in the `Config` map. Instead, this new field is where it should go.

pkg/apis/clusterregistry/types.go
perotinus: That's true. However, I expect this object to potentially grow, and the name is technically "metadata", so I think I'd prefer to leave it like this, since it describes the of this object.

pkg/apis/clusterregistry/types.go
perotinus: Good catch! I actually meant "to", not "so". Fixed.

PR #178, opened 2018-01-17
Support protocol buffers as a transport medium in the cluster registry

### LINKED ISSUE #106: Protobuf support
/sig multicluster

The cluster registry should support protocol buffers as an accepted content type. This may imply generating protocol buffers, the associated Go code, and then wiring up the API server properly.

Bazel may have issues with protocol buffer support, so this might be tricky.

perotinus:
/sig multicluster

Fixes #106.


font:
Also looks like this needs a rebase and retesting.

hack/update-codegen.sh
```diff
@@ -109,6 +110,18 @@ function generate_group() {
     --output-file-base zz_generated.conversion \
     "$4"
 
+  echo "generating protocol buffers for group ${GROUP_NAME} and version ${VERSION} at ${SCRIPT_BASE}/${CLIENT_PKG}"
+  bazel build \
+    //vendor/k8s.io/code-generator/cmd/go-to-protobuf/protoc-gen-gogo \
+    //third_party/protobuf:protoc
+
+  PATH="$(bazel info bazel-bin)/vendor/k8s.io/code-generator/cmd/go-to-protobuf/protoc-gen-gogo:$(bazel info bazel-bin)/third_party/protobuf:${PATH}" \
```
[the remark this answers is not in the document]
perotinus: This is an artifact of how the protocol buffer compiler works: it expects to be able to call plugins from the PATH. A bit of a pain, but able to be worked around like so.

Though, with the latest changes to `rules_go`, this has gotten a bit more complex.

perotinus:
@font Rebased, and I'll fix any additional test failures.

I removed the `third_party` directory, since I realized that a dependency is already pulling in the protobuf library into the bazel workspace. This might not be a great long-term solution, but at least for now I think we can rely on it, and it makes this PR simpler. The original reason for the `third_party` directory was that `dep` was not very good at pulling in the protobuf package, since it didn't all contain Go code. I added `protobuf` to `third_party` as a workaround (like k/k does, though they import significantly less of the repo), but changed that when I realized that I could use the version of `protobuf` from the bazel workspace.

perotinus:
Looks like `go-to-protobuf` calls `goimports` and `gofmt` in an `exec.Command`, assuming that it is in the path. This will require some extra thinking. I suppose the verify script could `go get` `gofmt` and `goimports` if they are not already in the `PATH`, but that's a bit of a pain

perotinus:
@font In any case, this should be ready to review; that infrastructure issue will need to be fixed before submission, but it shouldn't be a review blocker.

hack/update-codegen.sh
```diff
@@ -109,6 +110,22 @@ function generate_group() {
...
+  echo "generating protocol buffers for group ${GROUP_NAME} and version ${VERSION} at ${SCRIPT_BASE}/${CLIENT_PKG}"
+
+  # The generated go_binaries are not guaranteed to be in exactly the
+  # corresponding package in bazel-bin; there may be another
+  # architecture-specific subdirectory.
+  # The call to xargs trims leading whitespace.
+  PROTOC_GEN_GOGO_PATH="$(dirname "$(bazel build //vendor/k8s.io/code-generator/cmd/go-to-protobuf/protoc-gen-gogo:protoc-gen-gogo 2>&1 | grep bazel-bin | xargs)")"
+  PROTOC_PATH="$(dirname "$(bazel build @com_google_protobuf//:protoc 2>&1 | grep bazel-bin | xargs)")"
+
+  PATH="$(bazel info workspace)/${PROTOC_PATH}:$(bazel info workspace)/${PROTOC_GEN_GOGO_PATH}:${PATH}" \
```
font: Could we add a comment explaining that it's for the compiler to find the plugins?

Gopkg.toml
```diff
@@ -20,7 +22,10 @@ required = [
   "k8s.io/code-generator/cmd/set-gen",
 ]
 
-ignored = ["github.com/kubernetes/repo-infra/kazel"]
+ignored = [
+  "github.com/kubernetes/repo-infra/kazel",
+  "github.com/google/protobuf/*",
```
font: Is there a transitive dependency requiring `dep` to ignore?

perotinus: Hmm. I thought I had removed this, but I'll make sure it's actually gone. Perhaps it slipped into another commit as I was rebasing.

hack/update-codegen.sh
perotinus: Done.

perotinus:
@font I expect that some other dependency that's in the workspace requires the protobuf library, and so it's pulled in. My guess would be that `rules_go` depends on the protocol buffers library for the golang protobuf rules.

perotinus:
@font It looks like the `kubekins-e2e` image doesn't have `goimports` either. I suppose we could download it before running the test if it's not in the path.

perotinus:
@font OK, I fixed the issue with `goimports` by downloading it on each run of `hack/verify-all-gensrc` if it's not in the `PATH`. PTAL.

font:
@perotinus /lgtm

PR #180, opened 2018-01-19
Fix crinit standalone deployment

font:
This went into v0.0.2 with the recent `crinit` refactor, so we'll need to release v0.0.3.

/sig multicluster


perotinus:
/lgtm

Thanks, @font!

PR #181, opened 2018-01-19
Remove dependency on temporary GOPATH

### LINKED ISSUE #179: Remove dependency on temporary GOPATH in verify-go-src.sh
The existing `bazelbuild` image that is used to run the CI test is lacking a `GOPATH` environment setup that is required to run a non-bazel execute scenario requiring the Go toolchain, e.g. `go vet`, `go list`, etc. used by scripts called by `verify-go-src.sh`. This means we need to rely on creating our own temporary GOPATH as is currently being done in [`verify-go-src.sh`](https://github.com/kubernetes/cluster-registry/blob/master/hack/verify-go-src.sh).

Once https://github.com/kubernetes/test-infra/issues/6304 is resolved, we can simplify the `verify-go-src.sh` script to remove the dependency on creating our own temporary `GOPATH`.

/sig multicluster

font:
Fixes #179.

Now that we're relying on a different test image for the Prow job that runs this script, we can remove the dependency on a temporary `GOPATH` setup. This is because the new image contains the Go toolchain and a proper GOPATH setup prior to launching this script.

/sig multicluster

### DIFF
--- a/hack/verify-go-src.sh
+++ b/hack/verify-go-src.sh
@@ -22,18 +22,10 @@ set -euo pipefail
 
 SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "${SCRIPT_ROOT}/.." && pwd)"
-TMP_GOPATH="$(mktemp -d /tmp/gopathXXXXXXXX)"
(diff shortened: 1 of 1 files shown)


perotinus:
Thanks, @font!

Issue #48, opened 2017-10-25
crinit init should not fail if the resources already exist


perotinus:
Cloned from https://github.com/kubernetes/kubernetes/issues/50746.

`crinit init` should continue if related resources already exist in the host cluster; or it should provide a `--force` flag.

perotinus:
Also see https://github.com/kubernetes/kubernetes/issues/47159

G-Harmon:
I'm running into this. there are quite a few resources that I have to clean up manually when re-running crinit.
Here's the list of commands I needed to run in order to get my cluster cleaned up and ready to launch crinit again:

kc delete svc/first-cluster-reg -n clusterregistry ns/clusterregistry clusterroles/clusterregistry.k8s.io:apiserver clusterrolebindings/clusterregistry.k8s.io:apiserver  clusterrolebindings/clusterregistry.k8s.io:apiserver-auth-delegator clusterrolebindings/clusterregistry.k8s.io:extension-apiserver-authentication-reader apiservices/v1alpha1.clusterregistry.k8s.io; kc delete rolebindings/clusterregistry.k8s.io:extension-apiserver-authentication-reader -n kube-system

font:
Thanks @G-Harmon! Yep, we need to look at introducing a `delete` subcommand or eventually use helm, to make this simpler. I use the following:

`kubectl delete ns/clusterregistry clusterrole/clusterregistry.k8s.io:apiserver clusterrolebinding/clusterregistry.k8s.io:apiserver clusterrolebinding/clusterregistry.k8s.io:apiserver-auth-delegator; kubectl -n kube-system delete rolebinding/clusterregistry.k8s.io:extension-apiserver-authentication-reader apiservice/v1alpha1.clusterregistry.k8s.io`

font:
@jianzhangbjz While a `--force` option doesn't exist for the `init` subcommand, #192 adds the ability to delete the clusterregistry deployment and it has a `--force` option as well.

jianzhangbjz:
@font Oh, sorry, my misunderstanding. Got it, thanks!

perotinus:
Note that this issue is no longer applicable because of the CRD migration done in #225, which removes `crinit` and the standalone API server.

PR #192, opened 2018-02-03
Add aggregated and standalone delete sub-command with --force option

font:
- Separates out aggregated and standalone init and delete sub-commands into their own files.
- Adds delete sub-command to aggregated and standalone deployments.
- Adds a `--force` option for the delete sub-command.

This provides a fix for #48 by allowing the proper deletion of the cluster registry before proceeding with initializing it again.

Providing an equivalent `--force` flag for the `init` sub-command requires checking for an `AlreadyExists` object error status and if it exists, retrieve it to continue using that object. This retrieved object can potentially have been initialized with different options from the current run. That is, the already stored object may contain parameters that may preclude the option flags requested by the user during the latest run. This makes the solution not as clean IMO. So I'd like to propose moving forward with this solution for now.

/sig multicluster


pkg/crinit/common/common.go
```diff
@@ -451,3 +486,62 @@ func WaitSrvHealthy(cmdOut io.Writer, crClientset client.Interface) error {
...
+		_, err := fmt.Fprintln(cmdOut, "Cluster registry can be deleted (dry run)")
+		glog.V(4).Infof("Cluster registry can be deleted (dry run)")
+		return err
+	}
+
+	fmt.Fprint(cmdOut, "Waiting for the cluster registry to be deleted...")
+	glog.V(4).Info("Waiting for the cluster registry to be deleted")
+
+	var getErr error
+	err := wait.PollImmediate(waitInterval, waitTimeout, func() (bool, error) {
```
perotinus: Namespace deletion can take a very long time in my experience. We may need a longer timeout here, or some other way of indicating in the returned error that this is not necessarily unexpected.

Perhaps this command should continue and report some pseudo-success?

pkg/crinit/util/util.go
```diff
@@ -170,6 +170,54 @@ func UpdateKubeconfig(pathOptions *clientcmd.PathOptions, name, endpoint,
...
+
+	_, ok = kubeconfig.Clusters[name]
+	if !ok {
+		return fmt.Errorf("cannot delete cluster %s, not in %s", name, kubeconfigFile)
+	}
+
+	delete(kubeconfig.Contexts, name)
+	delete(kubeconfig.Clusters, name)
+
+	// Write the updated kubeconfig.
```
perotinus: What about the credentials?

font: Thanks for the reminder! I had actually thought I should delete the user authinfo as well, but since `kubectl config` doesn't have a way to view or delete just the user authinfo, I neglected to do it when I created this function and eventually forgot to revisit it. I've added it now.

pkg/crinit/aggregated/delete.go
```diff
@@ -0,0 +1,303 @@
...
+func runDelete(opts *aggregatedClusterRegistryOptions, cmdOut io.Writer,
+	hostClientset client.Interface, apiSvcClientset apiregclient.Interface,
+	pathOptions *clientcmd.PathOptions) error {
+
+	// Only necessary to delete the cluster registry namespace, the cluster
+	// registry API service, the RBAC objects that are not in the cluster
+	// registry namespace (those that are not removed by removing the
+	// namespace), and the kubeconfig entry.
+
+	err := common.DeleteKubeconfigEntry(cmdOut, pathOptions, opts.Name,
```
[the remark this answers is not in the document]
perotinus: That seems reasonable. The concern I would have is if a user has a `.kube/config` entry for some other cluster that this tool deletes, but I suppose that's not something we can adequately protect against.

The one thing I can think of as a protection is that we add info to the `Extensions` field that marks the context/cluster/authinfo as being created by this tool, and then check for that on deletion. I'm not sure that's worth the effort, and certainly not in this PR.

pkg/crinit/aggregated/delete.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	fmt.Fprintln(cmdOut, " done")
+	glog.V(4).Info("Successfully deleted cluster registry Kubernetes API Service")
+
+	return nil
+}
+
+// deleteAPIServiceObject deletes the cluster registry API Service object.
+func deleteAPIServiceObject(clientset apiregclient.Interface, name string,
```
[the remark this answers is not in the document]
perotinus: That's fair. I'm not sure that this pattern aids readability in the case where these methods don't do that much, but that's not something to change in this PR.

pkg/crinit/aggregated/delete.go
```diff
@@ -0,0 +1,302 @@
...
+		return err
+	}
+
+	glog.V(4).Info("Successfully deleted cluster role")
+
+	fmt.Fprintln(cmdOut, " done")
+	return nil
+}
+
+// deleteExtensionApiserverAuthenticationRoleBinding deletes the rolebinding
```
perotinus: nit: fix this name in the comment.

pkg/crinit/util/util.go
```diff
@@ -170,6 +170,86 @@ func UpdateKubeconfig(pathOptions *clientcmd.PathOptions, name, endpoint,
...
+
+	if dryRun {
+		return nil
+	}
+
+	// If we are not going to ignore errors, then return an error immediately
+	// on the first error encountered. If ignoring errors, then only output
+	// errors when verbose logging is enabled. If it turns out that all three:
+	// context, cluster, and authinfo were not found, then return an error as
+	// there is no point in updating the kubeconfig and the ignore errors flag
```
perotinus: What do you mean "the ignore errors flag will also be checked by the caller."?

pkg/crinit/util/util.go
```diff
@@ -170,6 +170,86 @@ func UpdateKubeconfig(pathOptions *clientcmd.PathOptions, name, endpoint,
...
+			return fmt.Errorf("cannot delete authinfo %s, not in %s", name, kubeconfigFile)
+		}
+		glog.V(4).Infof("cannot delete authinfo %s, not in %s", name, kubeconfigFile)
+		errCount++
+	} else {
+		delete(kubeconfig.AuthInfos, name)
+	}
+
+	if errCount == 3 {
+		return fmt.Errorf("cannot find %s context, cluster, and authinfo", name)
```
perotinus: Maybe update this error message to make it a bit more user friendly? Something like "Could not find any cluster registry context information (context, cluster or authinfo) in your kubeconfig."?

pkg/crinit/util/util.go
font: I updated the comment. Let me know if it is clear.

pkg/crinit/util/util.go
font: I like it - done.

pkg/crinit/options/options.go
```diff
@@ -103,6 +110,13 @@ func (o *SubcommandOptions) BindCommon(flags *pflag.FlagSet, defaultServerImage,
 		"Preferred port to use for the cluster registry API server NodePort service. Set to 0 to randomly assign a port. Valid only if '"+APIServerServiceTypeFlag+"=NodePort'.")
 }
 
+// BindCommonDelete adds the common options that are shared by the aggregated and
+// standalone delete sub-commands to the list of flags.
+func (o *SubcommandOptions) BindCommonDelete(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.IgnoreErrors, "ignore-errors", false,
+		"Run the command and ignore any errors encountered.")
```
perotinus: Maybe "Ignore errors while deleting resources. This will continue to delete all resources created by `init` even if an error occurs while deleting one of these resources.

font: I've updated the usage to a more concise version of your suggestion.

PR #193, opened 2018-02-08
Remove the internal version of the cluster API

perotinus:
/sig multicluster

After discussions with various people, I think that having the internal version doesn't provide enough benefit to be worth the maintenance cost. The one benefit that I know of for keeping the internal version is that it provides a standardized format for external versions to be converted into, but it seems like this purpose can be equally well served by an external type which other external types are converted into.

/cc @font @madhusudancs @pmorie @lavalamp @erictune


font:
I would say this LGTM, but I'm not sure how the internal version may or may not be used as I'm not really familiar with them. Do you know of any docs I could read?

perotinus:
@font I'm not sure there are any docs that describe how the internal version is used. In essence, from what I understand, the internal version is meant to be the version stored in etcd, which external versions are converted to and from before they are stored. So, if a user creates a `v1beta` object, that is converted into an `internal` object and stored; and if a user accesses a `v1beta2` object, the internal object is converted into a `v1beta2` object and returned to the user. This sort of separation between external and internal makes sense when the API doesn't have a strongly defined version, but since the Kubernetes APIs are versioned, I think using a named version (e.g., `v1beta1`) as the version in storage is OK for us for now; and it saves us the trouble of maintaining two different but almost equivalent object definitions.

font:
@perotinus Thanks for the explanation!

/lgtm

PR #194, opened 2018-02-08
Add cluster-registry API reference docs

font:
I plan to follow up with details on generating these docs in a future PR.

/sig multicluster

### DIFF
--- a/docs/api_reference.md
+++ b/docs/api_reference.md
@@ -1,0 +1,6 @@
+# cluster-registry API Reference
+
+## clusterregistry/v1alpha1
+
+-   [clusterregistry/v1alpha1 Operations](api-reference/clusterregistry/v1alpha1/operations.html)
+-   [clusterregistry/v1alpha1 Model Definitions](api-reference/clusterregistry/v1alpha1/definitions.html)
--- a/hack/update-api-reference-docs.sh
+++ b/hack/update-api-reference-docs.sh
@@ -1,0 +1,128 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
+#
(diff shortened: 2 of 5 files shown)


hack/update-api-reference-docs.sh
```diff
@@ -0,0 +1,128 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Generates updated api-reference docs from the latest OpenAPI spec for
+# clusterregistry apiserver. The docs are generated at docs/api-reference
+# Usage: ./update-api-reference-docs.sh <absolute output path>
+
+set -exuo pipefail
```
perotinus: Remove the `x`

docs/api_reference.md
```diff
@@ -0,0 +1,6 @@
+# cluster-registry API Reference
```
perotinus: Is this file generated? It seems like it could be generated by the script that creates the API documentation.

hack/update-api-reference-docs.sh
font: Good catch! You mean you don't want all the debug output too?! :wink:

docs/api_reference.md
font: It is not generated. It could be, but it should only ever change whenever there are new API versions. I don't really have a strong opinion here. Let me know if you'd prefer it to be generated and I can do a follow-up PR.

perotinus: Thanks! Given the infrequency of changes, I think it's probably not worth generating this right now.

PR #204, opened 2018-02-22
Fix clusterregistry.k8s.io:apiserver ClusterRole

mlowery:
I discovered this when I attempted to create a new `ClusterRoleBinding` referencing this `ClusterRole`. Until I changed from `cluster` to `clusters`, it didn't work.

Before:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app: clusterregistry
  name: clusterregistry.k8s.io:apiserver
rules:
- apiGroups:
  - clusterregistry.k8s.io
  resources:
  - cluster
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
```

After (only change is `clusters` under `resources`):

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app: clusterregistry
  name: clusterregistry.k8s.io:apiserver
rules:
- apiGroups:
  - clusterregistry.k8s.io
  resources:
  - clusters
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
```

This brings up another question. What is the purpose of this `ClusterRole`? It's bound to the `ServiceAccount` in the `ClusterRoleBinding` of the same name but I'm not familiar with when that would be used. Any instruction here would be appreciated.

/sig multicluster

### DIFF
--- a/pkg/crinit/aggregated/aggregated.go
+++ b/pkg/crinit/aggregated/aggregated.go
@@ -53,7 +53,7 @@ var (
 
 	// The list of our cluster registry API resources to which the rule applies
 	// in the cluster role object.
-	clusterRoleResources = []string{"cluster"}
+	clusterRoleResources = []string{"clusters"}
 
 	// The list of verbs that apply to our cluster registry API resources.
 	clusterRoleVerbs = []string{"get", "list", "watch", "create", "update", "patch", "delete"}


font:
> This brings up another question. What is the purpose of this ClusterRole? It's bound to the ServiceAccount in the ClusterRoleBinding of the same name but I'm not familiar with when that would be used. Any instruction here would be appreciated.

@mlowery Essentially, the cluster role binding links the service account and cluster role to provide the RBAC permissions to allow all the necessary operations on the `clusters` resource as shown here:
```
  "kind": "APIResourceList",                                               
  "apiVersion": "v1",                                                                                            
  "groupVersion": "clusterregistry.k8s.io/v1alpha1",                                            
  "resources": [                                                                 
    {                                                                               
      "name": "clusters",
      "singularName": "",
      "namespaced": false,
      "kind": "Cluster",
      "verbs": [
        "create",
        "delete",
        "deletecollection",
        "get",
        "list",
        "patch",
        "update",
        "watch"
      ]
    }
  ]

```

The guidance for this is found in the [k8s docs here](https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/#setup-an-extension-api-server-to-work-with-the-aggregation-layer).

mlowery:
@font: Thanks for the merge and the info. I think what I mean is, given that it wasn't working before (for me at least), does that mean that this `ClusterRole` wasn't being used and therefore unnecessary?

I agree that someone at some point has to grant access to cluster objects to someone. But the grant here (via the `ClusterRoleBinding` of the same name) is to the service account. And I'm not sure when the incoming identity would ever be that.

font:
@mlowery I've created #207 to remove these objects as I don't think they're needed since we won't have a controller that needs permissions to operate on our cluster resource.

PR #205, opened 2018-02-22
Add support to crinit for running cluster registry outside of aggregator

### LINKED ISSUE #203: crinit should support running outside of aggregator in aggregated mode
Today, when running the `aggregated` subcommand, `crinit` will create the extension apiserver in the aggregator (since it only takes one context `host-cluster-context`). In some circumstances, an operator may want to run cluster registry outside of the aggregator. I have customized `crinit` to support this. Is this a feature that you would be interested in? A summary of changes:

* takes additional kubeconfig context `agg-context` (defaults to value given in `host-cluster-context` for backward compatibility)
* passes `request-header-*` to cluster registry apiserver (since it cannot get these from the `extension-apiserver-authentication` configmap at runtime through in-cluster client)
* passes `{authentication,authorization}-kubeconfig` to cluster registry apiserver (since it cannot use in-cluster client) and mounts related secret (after serviceaccount created, fetches token from its related secret in aggregator and uses that as secret)
* takes additional flag `canonical-name` which becomes the `externalName` value in a service of type `ExternalName` which lives in the aggregator and points to the "real" service in the host cluster
* skips some steps that are not needed when running outside the aggregator (e.g. creating a rolebinding for reading `extension-apiserver-authentication` configmap)

/sig multicluster

mlowery:
Closes #203

/sig multicluster

### DIFF
--- a/pkg/crinit/aggregated/aggregated.go
+++ b/pkg/crinit/aggregated/aggregated.go
@@ -19,6 +19,7 @@ package aggregated
 import (
 	"io"
 	"strings"
+	"time"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
@@ -72,15 +73,28 @@ var (
 	extensionAPIServerRBName = v1alpha1.GroupName + ":extension-apiserver-authentication-reader"
 )
 
+const (
+	serviceAccountSecretTimeout = 30 * time.Second
+	canonicalNameFlag           = "canonical-name"
+)
+
 type aggregatedClusterRegistryOptions struct {
 	options.SubcommandOptions
 	apiServerServiceTypeString string
+	AggContext                 string
+	CanonicalName              string
 }
 
 func (o *aggregatedClusterRegistryOptions) Bind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.apiServerServiceTypeString, options.APIServerServiceTypeFlag,
 		string(v1.ServiceTypeNodePort),
 		"The type of service to create for the cluster registry. Options: 'LoadBalancer', 'NodePort'.")
+	flags.StringVar(&o.AggContext, "agg-context",
+		"",
+		"Context of the aggregator. Defaults to same as "+options.HostClusterContextFlag+".")
+	flags.StringVar(&o.CanonicalName, canonicalNameFlag,
+		"",
+		"Canonical name of service running cluster registry (used when running outside of aggregator.")
 }
 
 // NewCmdAggregated defines the `aggregated` command with `init` and `delete`
--- a/pkg/crinit/aggregated/init.go
+++ b/pkg/crinit/aggregated/init.go
@@ -20,17 +20,21 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 	"k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/util/cert"
+	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/cluster-registry/pkg/apis/clusterregistry/v1alpha1"
 	crclient "k8s.io/cluster-registry/pkg/client/clientset_generated/clientset"
 	"k8s.io/cluster-registry/pkg/crinit/common"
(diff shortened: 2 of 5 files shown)


pkg/crinit/aggregated/aggregated.go
```diff
@@ -72,15 +73,28 @@ var (
 	extensionAPIServerRBName = v1alpha1.GroupName + ":extension-apiserver-authentication-reader"
 )
 
+const (
+	serviceAccountSecretTimeout = 30 * time.Second
```
perotinus: serviceAccountSecretCreationTimeout

? Is it a creation timeout?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -72,15 +73,28 @@ var (
...
+	canonicalNameFlag           = "canonical-name"
+)
+
 type aggregatedClusterRegistryOptions struct {
 	options.SubcommandOptions
 	apiServerServiceTypeString string
+	AggContext                 string
```
perotinus: Can you document these?

pkg/crinit/aggregated/aggregated.go
```diff
@@ -72,15 +73,28 @@ var (
...
+	CanonicalName              string
 }
 
 func (o *aggregatedClusterRegistryOptions) Bind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.apiServerServiceTypeString, options.APIServerServiceTypeFlag,
 		string(v1.ServiceTypeNodePort),
 		"The type of service to create for the cluster registry. Options: 'LoadBalancer', 'NodePort'.")
+	flags.StringVar(&o.AggContext, "agg-context",
```
perotinus: `aggregator-context`

pkg/crinit/aggregated/init.go
```diff
@@ -85,18 +89,25 @@ func newSubCmdInit(cmdOut io.Writer, pathOptions *clientcmd.PathOptions,
...
 				glog.Fatalf("error: %v", err)
 			}
-
 			hostClientset, err := client.NewForConfig(hostConfig)
 			if err != nil {
 				glog.Fatalf("error: %v", err)
 			}
 
-			apiServiceClientset, err := apiregclient.NewForConfig(hostConfig)
+			aggConfig, err := util.GetClientConfig(pathOptions, opts.AggContext, opts.Kubeconfig).ClientConfig()
```
perotinus: `aggregatorConfig`

Since there could be some confusion between "aggregator" and "aggregated", I'd prefer to spell this out in full wherever it appears.

pkg/crinit/aggregated/init.go
```diff
@@ -111,15 +122,29 @@ func newSubCmdInit(cmdOut io.Writer, pathOptions *clientcmd.PathOptions,
 }
 
 // validateOptions ensures that options are valid.
-func validateOptions(opts *aggregatedClusterRegistryOptions) error {
+func validateOptions(opts *aggregatedClusterRegistryOptions, cmdOut io.Writer) error {
 	opts.APIServerServiceType = v1.ServiceType(opts.apiServerServiceTypeString)
+	if len(opts.AggContext) == 0 {
+		glog.Warningf(
```
perotinus: I'm not sure this is worth a warning: it seems like a common use case that people will want to aggregate to the same context that the pod is running in.

pkg/crinit/aggregated/init.go
```diff
@@ -111,15 +122,29 @@ func newSubCmdInit(cmdOut io.Writer, pathOptions *clientcmd.PathOptions,
...
+			"Using context %q for aggregator since no dedicated context specified", opts.Host)
+		fmt.Fprintln(cmdOut, "WARNING: Using context %q for aggregator since no dedicated context specified")
+		opts.AggContext = opts.Host
```
perotinus: I think this should be moved elsewhere: this isn't a validation step, but rather a defaulting step.

pkg/crinit/aggregated/init.go
```diff
@@ -111,15 +122,29 @@ func newSubCmdInit(cmdOut io.Writer, pathOptions *clientcmd.PathOptions,
...
+	}
+	if !runningInAggregator(opts) {
+		if len(opts.CanonicalName) == 0 {
+			return fmt.Errorf("%s is required when running outside of aggregator", canonicalNameFlag)
+		}
+	}
 	return opts.ValidateCommonOptions()
 }
 
+func runningInAggregator(opts *aggregatedClusterRegistryOptions) bool {
```
perotinus: Can you document this function?

pkg/crinit/aggregated/init.go
```diff
@@ -131,6 +156,18 @@ func runInit(opts *aggregatedClusterRegistryOptions, cmdOut io.Writer,
 		return err
 	}
 
+	if !runningInAggregator(opts) {
+		// separate apiserver when not running in aggregator; need to create namespace
```
perotinus: Can you clarify this comment a bit, maybe adding a bit more detail? Why do you need to create a namespace?

pkg/crinit/aggregated/init.go
perotinus: I think the logic might be simpler to understand if you reused the hostClientset if the aggregator-context flag is not set, rather than creating a new clientset that points to the same host.

pkg/crinit/common/common.go
```diff
@@ -47,7 +47,8 @@ const (
 	// Set the secure port to 8443 to avoid requiring root privileges
 	// to bind to port < 1000.  The apiserver's service will still
 	// expose on port 443.
-	apiServerSecurePort = 8443
+	apiServerSecurePort     = 8443
+	KubeconfigSecretDataKey = "kubeconfig"
```
perotinus: Can you document this?

perotinus:
@mlowery I think this PR is no longer necessary, in light of #225 which removed `crinit` entirely.

Issue #210, opened 2018-03-01
Add columns to `kubectl get clusters`


pmorie:
Currently, the cluster registry does not send OpenAPI column information for use with kubectl. We should begin sending this so that users get more information when they do `kubectl get clusters`.

/sig multicluster

pmorie:
@onyiny-ang

PR #212, opened 2018-03-12
[WIP] Adding 'kubectl' column Info using table convertor

onyiny-ang:
/sig multicluster

Addresses #210


onyiny-ang:
Closing this for now as it doesn't seem like it can/should be implemented any time soon.

PR #225, opened 2018-05-08
Migrate the cluster registry to use a CRD

perotinus:
This PR has many commits. I think it will be easier to review commit-by-commit rather than reviewing the entire PR at once.

As a follow-up to this, I (or someone else) will need to:
- [ ] switch to a namespaced object
- [ ] remove the unnecessary test jobs
- [ ] figure out what needs to be done with `LICENSES`
- [ ] create a release process for the CRD
- [ ] figure out if/how to write a test to verify that the `kubebuilder` generated code is up-to-date
- [ ] close bugs made irrelevant by this PR
- [ ] validation

/sig multicluster

cc @font @pmorie @madhusudancs


perotinus:
The test failures are expected: `bazel` isn't used anymore in the repo, and the code generation does not have a verification script working yet.

perotinus:
@pmorie Yes, I do: I've removed all the bazel infrastructure, as well as the code generation scripts. I think I will submit a PR to remove the bazel jobs when this PR is approved, just before merging; and I can add a dummy script that makes the `pull-cluster-registry-verify-gensrc` pass, before later adding support for verifying the `kubebuilder` generation.

pmorie:
@perotinus that sounds like a sensible approach.

perotinus:
@pmorie Any objections to this going in?

hack/go-tools/verify-govet.sh
```diff
@@ -18,5 +18,5 @@ set -euo pipefail
 SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "${SCRIPT_ROOT}/../.." && pwd)"
 pushd ${REPO_ROOT} > /dev/null
-go vet -v $(go list ./... | grep -v /vendor/)
+go vet -v $(go list ./... | grep -v /vendor/ | grep -v /client/)
```
font: Consider using 1 grep command: `grep -Ev '/(vendor|client)/'`

pkg/apis/clusterregistry/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -82,41 +82,41 @@ var (
 				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
 					Type: "object",
 					Properties: map[string]v1beta1.JSONSchemaProps{
-						"apiVersion": v1beta1.JSONSchemaProps{
+						"apiVersion": {
```
font: It seems like `kubebuilder` needs to be optimized to produce go-vetted code in this case.

hack/verify-all-gensrc.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+# This script runs all of the Go source verification scripts inside of the
+# ./hack/go-tools directory. The success or failure of each script is outputted
+# in green or red colored text, respectively. If any script fails, an error is
+# returned, otherwise returns 0.
+
+set -euo pipefail
+
+# TODO: Run `kubebuilder generate` and verify that nothing has changed.
```
font: We'll also need to run `go vet` as well.

font:
The only other thing we need to add to the list of follow-ups is validation for this CRD.

/lgtm

pkg/apis/clusterregistry/v1alpha1/zz_generated.kubebuilder.go
perotinus: Yes! Filed https://github.com/kubernetes-sigs/kubebuilder/issues/171

docs/userguide.md
```diff
@@ -2,92 +2,44 @@
...
+generated client libraries, and supports having controllers watch for updates.
 
-## Deploying a cluster registry
+## Setting up a cluster registry
 
-The `crinit` tool allows you to deploy a cluster registry into an existing
-Kubernetes cluster. Run the following commands to download the latest nightly
-build of the tool:
+The cluster registry API is defined as a [Kubernetes custom resource
+definition](https://kubernetes.io/docs/concepts/api-extension/custom-resources/).
```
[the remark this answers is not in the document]
perotinus: Good call, done.

hack/go-tools/verify-govet.sh
perotinus: Nice! Fixed.

hack/verify-all-gensrc.sh
perotinus: Added a note.

pkg/inject/inject.go
```diff
@@ -0,0 +1,29 @@
+package inject
```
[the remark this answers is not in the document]
font: Ah, ok. Do we need this code then since we don't need a controller?

font:
Thanks!
/lgtm

Issue #229, opened 2018-05-17
Follow-up tasks for CRD conversion


perotinus:
/sig multicluster

Follow-up tasks from #225, migrated here for easier tracking
- [x] switch to a namespaced object (#228)
- [x] remove the unnecessary test jobs
- [x] figure out what needs to be done with `LICENSES`
- [ ] create a release process for the CRD
- [ ] figure out if/how to write a test to verify that the `kubebuilder` generated code is up-to-date
- [ ] run the integration tests in `/test/...` before each PR
- [x] close bugs made irrelevant by this PR
- [ ] validation
- [x] look at removing `pkg/inject` if it's not necessary.
- [x] fix issue with cluster CRD `additionalPermissions` field (#231)

PR #237, opened 2018-05-29
Remove unnecessary code from pkg/inject, and clarify why that package and pkg/controller exist.

perotinus:
/sig multicluster

Addresses one of the follow-up items in #229.

/assign @font @madhusudancs @pmorie


font:
Is there no way to disable the creation of the controller generated code? It looks like `kubebuilder` offers a `--controller` flag to make generating the controller code optional when creating a resource as in:
```bash
kubebuilder create resource --controller=false ...
```

perotinus:
@font There is some code in `kubebuilder docs` that specifically references in a non-switchable way `pkg/controller`: https://github.com/kubernetes-sigs/kubebuilder/blob/561e4da021cdee17759d74580829ab1e9a6eb6cb/cmd/kubebuilder/docs/gen.go#L25

Filed https://github.com/kubernetes-sigs/kubebuilder/issues/249 to see if kubebuilder want to fix this internally.

font:
Ah ok, thanks! Yes, kubebuilder should allow the controller to be optional when generating docs.

/lgtm

PR #250, opened 2018-06-28
Use `gofmt -s -w` to format the file.

gyliu513:
`gofmt -s` will not write result to source file.

/sig multicluster

### DIFF
--- a/hack/go-tools/verify-gofmt.sh
+++ b/hack/go-tools/verify-gofmt.sh
@@ -27,12 +27,12 @@ find_files() {
     \) -name '*.go'
 }
 
-GOFMT="gofmt -s -w"
+GOFMT="gofmt -s"
 bad_files=$(find_files | xargs $GOFMT -l)
 popd > /dev/null
 
 if [[ -n "${bad_files}" ]]; then
-  echo "!!! '$GOFMT' needs to be run on the following files: "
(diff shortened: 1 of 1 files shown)


hack/go-tools/verify-gofmt.sh
```diff
@@ -27,7 +27,7 @@ find_files() {
     \) -name '*.go'
 }
 
-GOFMT="gofmt -s"
+GOFMT="gofmt -s -w"
```
perotinus: I'm not sure this is quite what we want to do: this script is run during a Prow job, and having it do the writes in place does not help us there. Does `gofmt -s -w` also write out the names of the files that it simplifies?

I think, to add this line, it would make sense to add an `update-gofmt.sh` script that involved the `gofmt` command correctly.

gyliu513: @perotinus I only updated the promoted message in the new pr, can you help check? The major improvement that I want to do is to give developer a good message for the next action.

perotinus: @gyliu513 Sure, this looks good to me. Thank yoU!

Can you merge the two commits so that there isn't noise in the change?

gyliu513: seems squash does not work, created another PR at https://github.com/kubernetes/cluster-registry/pull/251

PR #252, opened 2018-07-02
Added status to cluster yaml template.

gyliu513:
/sig multicluster

/cc @perotinus


perotinus:
@gyliu513 Thank you! Does this also need to be added to the example files, or do they work as-is?

- https://github.com/kubernetes/cluster-registry/blob/master/examples/cluster.yaml
- https://github.com/kubernetes/cluster-registry/blob/master/examples/cluster-two.yaml

gyliu513:
Thanks @perotinus , yes, we should also update the samples as you mentioned, new commits updated.

Issue #253, opened 2018-07-11
Link to YAML for CRD leads to 404 on website


thisisshi:
https://kubernetes.github.io/cluster-registry/userguide.html

the link "YAML for the CRD" leads to https://kubernetes.github.io/cluster-registry-crd.yaml which is a 404

/sig multicluster

PR #254, opened 2018-07-11
fix crd yaml path in website, previously was 404

thisisshi:
Related to #253 

/sig multicluster


pmorie:
Thanks for this @thisisshi

PR #261, opened 2018-09-25
(Request for comments) Add ClusterCredentials API

pmorie:
Adds a `ClusterCredentials` API type that provides a kubeconfig to communicate with a cluster in the cluster registry. Submitting for comments per our recent discussion in the SIG.

/sig multicluster

### DIFF
--- a/pkg/apis/clusterregistry/v1alpha1/types.go
+++ b/pkg/apis/clusterregistry/v1alpha1/types.go
@@ -181,6 +181,10 @@ type ClusterCondition struct {
 // within the cluster hosting the ClusterCredentials resource - to access the
 // cluster's healthz endpoint.
 //
+// ClusterCredentials is meant to enable creation of controllers that interact
+// with remote clusters and/or software that generates kubeconfig files that are
+// usable with remote clusters.
+//
 // +k8s:openapi-gen=true
 // +kubebuilder:resource:path=clustercredentials
 // +kubebuilder:subresource:status
@@ -201,17 +205,31 @@ type ClusterCredentialsSpec struct {
 	ClusterRef v1.ObjectReference `json:"clusterRef,omitempty"`
 
 	// SecretRef is a reference to the secret in the namespace that contains a
-	// kubeconfig with credentials access the referenced cluster. The secret
-	// must have a key named "kubeconfig" that contains a file formatted as a
-	// proper subset of the kubeconfig file format. The data must contain a
-	// kubeconfig file with only the following fields:
+	// kubeconfig user with authentication information to use to access the
+	// referenced cluster. The secret must have a key named "authInfo" that
+	// contains a serialized kubeconfig AuthInfo.
 	//
-	// - a 'users' section
-	// - a single, valid entry in the 'users' section containing the auth
-	//   information to use the referenced Cluster.
+	// The AuthInfo type is defined here:
 	//
-	// TODO: explain which auth methods are supported
-	// TODO: explain kubeconfig version
+	// https://github.com/kubernetes/client-go/blob/v9.0.0/tools/clientcmd/api/v1/types.go
+	//
+	// Not all fields of AuthInfo are supported; unsupported fields are:
+	//
+	// - auth-provider
+	// - exec
+	// - extensions
+	//
+	// Note, these fields are currently unsupported to make the semantics of
+	// this API easier to reason about; they may be allowed in the future (or
+	// even later in the lifespan of this PR).
+	//
+	// TODO: firm up relationship of clientcmdapi to cluster-registry re:
+	// version
+	//
+	// TODO: determine whether this is the right API to hold information
+	// like:
+	// - CA bundle
+	// - skip TLS verify
 	//
 	// This can be left empty if the cluster allows insecure access.
 	// +optional
--- a/pkg/apis/clusterregistry/v1alpha1/types.go
+++ b/pkg/apis/clusterregistry/v1alpha1/types.go
@@ -155,6 +155,10 @@ type ClusterCondition struct {
 	// +optional
 	LastHeartbeatTime metav1.Time `json:"lastHeartbeatTime,omitempty" protobuf:"bytes,3,opt,name=lastHeartbeatTime"`
 
+	// Last time the condition was checked.
+	// +optional
+	LastProbeTime metav1.Time `json:"lastProbeTime,omitempty"`
+
 	// LastTransitionTime is the last time the condition changed from one status to another.
 	// +optional
 	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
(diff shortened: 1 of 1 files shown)


pmorie:
For the record, and in case it's not clear - there will be an accompanying controller that updates the status of `ClusterConnection`. Once we agree on the API I will add that to this PR.

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,81 @@ type ClusterCondition struct {
...
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterConnection represents a connection to a Cluster using a Kubeconfig.
+//
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=clusterconnections
+// +kubebuilder:subresource:status
+type ClusterConnection struct {
```
pmorie: Note: I do not like this name but didn't want to spend a lot of time naming before a group discussion

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+	Status ClusterCredentialsStatus `json:"status,omitempty"`
+}
+
+// ClusterCredentialsSpec defines the desired state of a ClusterCredentials.
+type ClusterCredentialsSpec struct {
+	// ClusterRef is a reference to the Cluster to access.
+	ClusterRef v1.LocalObjectReference `json:"clusterRef,omitempty"`
+
+	// SecretRef is a reference to the secret in the namespace that contains a
+	// kubeconfig with credentials access the referenced cluster. The secret
```
font: s/credentials access/credentials to access/

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+	// Status is the status of the condition. One of True, False, Unknown.
+	Status v1.ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status,casttype=ConditionStatus"`
+
+	// LastHeartbeatTime is the last time this condition was updated.
+	// +optional
+	LastHeartbeatTime metav1.Time `json:"lastHeartbeatTime,omitempty" protobuf:"bytes,3,opt,name=lastHeartbeatTime"`
+
+	// Last time the condition was checked.
+	// +optional
+	LastProbeTime metav1.Time `json:"lastProbeTime,omitempty"`
```
font: Missing some struct tags like the other fields in this struct.

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+}
+
+// ClusterCredentialsConditionType marks the kind of ClusterCredentials condition
+// being reported.
+type ClusterCredentialsConditionType string
+
+const (
+	// ClusterCredentialsHealthy represents that a ClusterCredentials is healthy
+	// as determined by the Cluster's healthz endpoint.
+	ClusterCredentialsHealthy ClusterConditionType = "Healthy"
```
font: s/ClusterConditionType/ClusterCredentialsConditionType/

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+	Reason string `json:"reason,omitempty" protobuf:"bytes,5,opt,name=reason"`
+
+	// Message is a human-readable message indicating details about the last status change.
+	// +optional
+	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
+}
+
+// ClusterCredentialsConditionType marks the kind of ClusterCredentials condition
+// being reported.
+type ClusterCredentialsConditionType string
```
font: It seems like there may be some overlap with `ClusterConditionType`. Any reason not to use it instead by just adding a new `ClusterCredentialsHealthy` const to it?

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+	// TODO: explain kubeconfig version
+	//
+	// This can be left empty if the cluster allows insecure access.
+	// +optional
+	SecretRef *apiv1.LocalObjectReference `json:"secretRef,omitempty"`
+}
+
+// ClusterCredentialsStatus contains information about the current status of a
+// ClusterCredentials.
+type ClusterCredentialsStatus struct {
```
pmorie: Note: I removed `availabilityZone` and `region` from the status of `FederatedCluster` when I used that as inspiration for this type. Do we think those concepts should be present in this API?

jonmoter:
It would be useful to have requirements or use-cases for this feature, to know the scope of what it is trying to solve. Since I'm not clear what this API is attempting to model or accomplish, it's hard for me to give feedback on how well it's succeeding.

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+}
+
+// ClusterCredentialsSpec defines the desired state of a ClusterCredentials.
+type ClusterCredentialsSpec struct {
+	// ClusterRef is a reference to the Cluster to access.
+	ClusterRef v1.LocalObjectReference `json:"clusterRef,omitempty"`
+
+	// SecretRef is a reference to the secret in the namespace that contains a
+	// kubeconfig with credentials access the referenced cluster. The secret
+	// must have a key named "kubeconfig" that contains a valid kubeconfig.
```
jonmoter: It seems strange to me to have an entire kubeconfig file in an object that is called "Credentials". A kubeconfig file includes:

* `clusters` - an array of connection information (endpoints, TLS details)
* `users` - an array of credentials used for AuthN (tokens, client certs, instructions to execute CLI binaries, etc.)
* `contexts` - a combination of `cluster` + `user`, optionally with a namespace
* `current-context` - (optional) currently selected context

Of those, only entries in the `users` array make sense from the perspective of Credentials.

If the kubeconfig file has one or more `clusters` defined in it, should I client tool use one of those endpoints, or the endpoints specified in the `ClusterSpec.KubernetesAPIEndpoints` object?

I think it would make more sense to have a secret that contained a single `user` object in the kubeconfig file. Then it would be possible to construct a kubeconfig file using a combination of the `KubernetesAPIEndpoints` objects and `ClusterCredentialsSpec` objects.

quinton-hoole-2:
> It would be useful to have requirements or use-cases for this feature, to know the scope of what it is trying to solve. Since I'm not clear what this API is attempting to model or accomplish, it's hard for me to give feedback on how well it's succeeding.

Good feedback @jonmoter .  Here is an excerpt from previous meeting notes where it was discussed, to provide further context:

```
perotinus] Moving clusterregistry.Cluster to beta; minor update to cluster registry.
Quinton: How do people use cluster registry without auth-info (which seems to currently be an unused field in cluster-registry).
Paul: The field is just to indicate what/how the auth should be done. Personally does not feel that cluster object need to have the auth info field.
Quinton: How does cluster registry become useful without the authentication information for a user who discovers the clusters in cluster registry?
Jonathan: The initial idea of auth field in cluster registry against a cluster was to provide enough info to the user to figure out how to get this auth info.
Quinton: Ideally we should be able to do away with the FederatedCluster resource in federation and enable cluster registry to provide more useful information rather than just the API endpoint of the cluster.
Need to have more discussions on this before further moving it to beta.
```

What wasn't captured in the notes, but from my memory, is that perotinus and others made points that:

1. there are currently no users of cluster registry other than Federation-v2 (that he was aware of)
2. that it was therefore challenging to validate (or otherwise) whether the cluster registry API was "acceptable enough" to be firmed up to beta
3. that Federation-v2 had to build an addon, of approximately similar complexity, to configure cluster credentials, in order to actually use cluster registry.  So in practise the cluster registry by itself did not save Federation-v2 much time over rolling it's own.  Similar reasoning may also explain why nobody else was using cluster registry.

Hence it was decide to explore what would need to be added to cluster registry w.r.t. cluster credentials to obviate the need for the addons in Federation-v2, hopefully to pave the way for other users.

perotinus:
Sorry for not taking a look at this sooner!

Some overall thoughts:
- I like the overall concept that allows a 1-N relationship between clusters and user credentials
- using the `kubeconfig` like this concerns me: AFAICT it's meant to be for one user on one machine. I understand the appeal of making it easier to bootstrap with an existing file, but the number of fields that are not needed in this object, some of which are duplicated with the Cluster object itself, give me pause
- the `kubeconfig` file can farm out to custom commands to get tokens, which may not work from within a clusters
- there are changes coming down the pike (expiration being a major one that comes to mind) to Kubernetes service accounts that will make it less appealing to use their tokens

pkg/apis/clusterregistry/v1alpha1/types.go
```diff
@@ -167,3 +171,87 @@ type ClusterCondition struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ClusterCredentialsSpec   `json:"spec,omitempty"`
+	Status ClusterCredentialsStatus `json:"status,omitempty"`
+}
+
+// ClusterCredentialsSpec defines the desired state of a ClusterCredentials.
+type ClusterCredentialsSpec struct {
+	// ClusterRef is a reference to the Cluster to access.
+	ClusterRef v1.LocalObjectReference `json:"clusterRef,omitempty"`
```
perotinus: If this were an object reference, it would be easier to have a cluster registry with a public list of clusters (in, say, `kube-multicluster-public`) and have each user/team have their own ClusterConnection in their own namespace. That would make access control easier to reason about.

pmorie: That's a very good point, thanks for the feedback. I will add a distinct option for a namespace-scoped reference

pkg/apis/clusterregistry/v1alpha1/types.go
pmorie: That's a good suggestion. I'll incorporate this into the doc for this field (I don't believe it should change the API as currently written from a structural standpoint).

quinton-hoole-2:
@pmorie As discussed in person, I think that particular name is a bad idea.  We're not talking about health checks here, we're talking about authenticating against a cluster. Personally I don't mind "ClusterCredentials" because that's exactly what they are.  But feel free to find another name that's not confusing or misleading (which adding "HealthChecked" would be).

pkg/apis/clusterregistry/v1alpha1/types.go
shashidharatd: IMHO, labels should be good enough like the labels on node. `failure-domain.beta.kubernetes.io/region` and `failure-domain.beta.kubernetes.io/zone`. we could use the same labels on cluster object.

pmorie:
@sohankunkerkar is going to pick this up and open a new PR. Apologies for the latency!

pmorie:
I'm honestly not sure whether this PR should remain open at this point. Since I opened it, it's come to light that the way we were using the cluster-registry API in the kubefed project allowed escalation: if you have write access to the `Cluster` resource, you can effectively steal the secrets associated with a `ClusterCredentials` (or `KubeFedCluster` in the case of kubefed). I'm not sure I want to continue this PR since it would build on that usage pattern.

quinton-hoole:
@pmorie It's not immediately clear how having write access to Cluster allows stealing of Secrets referred to by ClusterCredentials.  Could you clarify, or provide a link to an explanation?

Intiutively it seems that RBAC rules on the Secret should control read access to the secrets.  But presumably you've discovered a hole in that?

PR #269, opened 2018-12-07
Fixing broken CRD definition

### LINKED ISSUE #255: CRD definition is broken on kube 1.11.1 cluster
/sig multicluster
 when applying the crd config to a kube 1.11.1 cluster, it will error out with this message:
```
$ kubectl apply -f cluster-registry-crd.yaml
error: error validating "cluster-registry-crd.yaml": error validating data: [ValidationError(CustomResourceDefinition.status): missing required field "conditions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus, ValidationError(CustomResourceDefinition.status): missing required field "storedVersions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus]; if you choose to ignore these errors, turn validation off with --validate=false
```

Currently we worked around the issue by updating CRD with this change although we are not quite sure if that's the right way to fix it:
```
-  conditions: null
+  conditions: []
+  storedVersions: []
```

poothia:
fixes #255 

reference https://github.com/kubernetes-sigs/federation-v2/pull/369

/sig multicluster

### DIFF
--- a/cluster-registry-crd.yaml
+++ b/cluster-registry-crd.yaml
@@ -86,8 +86,3 @@ spec:
               type: array
           type: object
   version: v1alpha1
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: null


normanjoyner:
This looks like the right approach to me. Any thoughts @madhusudancs / @perotinus ?

perotinus:
/lgtm
/approve

Apologies, I have not been looking at Github notifications recently.

PR #275, opened 2019-06-18
Update to k8s-ci-robot's command list

chases2:
The previous link was broken

/sig multicluster


chases2:
I have no idea how this change and that failure might be related.

/retest

poothia:
lgtm

